PR #46, opened 2025-08-16
Unify Dockerfile for GPU (nvidia) and CPU, use runtime CUDA images instead of devel

HeroBrine1st:
Also refactor Dockerfile so that it *builds* not redownloads everything on each change.

### TODO

- [ ] Find out lowest CUDA version (looks like 12.2 is a good baseline to start searching as it is supports 525 drivers). Determine which variant it needs (resulting image is more than 3.5GB!)
- [x] Update github workflows
- [ ] Test on multiple other configurations
- [ ] Update compose files if needed
- ~~TF_FORCE_GPU_ALLOW_GROWTH is questionable. Currently reallocates memory for every task on my machine (idk whether it does that by default)~~ it does not in both cases if done right

### How to build

Builds to CPU out of the box. To build for nvidia:

```yaml
services:
  <service_name>:
    build:
      context: ./AudioMuse-AI
      args:
        BASE_IMAGE: nvidia/cuda:12.8.1-cudnn-runtime-ubuntu22.04
```

~~Yes, it will switch tensorflow package~~ isn't needed, works out of the box with CUDA

---

For reference, my deploy section (for both flask and worker):

```yaml
    deploy:
      resources:
        reservations:
          devices:
           - driver: cdi
             device_ids:
                - nvidia.com/gpu=all
             capabilities: ["gpu"]
```

I have custom stripped-down version of NixOS nvidia module on my server - https://github.com/HeroBrine1st/flake/blob/3d03cb1b6ed882a9dd76f247723883354534fffb/modules/optionals/nvidia-ctk-isolated.nix

Tested on tag v0.6.4-beta


NeptuneHub:
**Background:** I actually don't have GPU to test nvdia stuff, my actual configuration use k3s on 4 low energy / low cost node to parallelize the work on multiple CPU more than using GPU. So meanwhile I don't have a way to test this I can only trust the community.

Say that actually also @ravenscroftj is working on the nvidia support, and his code is actually merged in main branch, the container with tag `latest-nvidia` actually use it.

My suggestions is that you could work together in a solution that work good for both.

**Definition of good:** It enable nvidia GPU + don't introduce error + work faster than using the single CPU

Meanwhile really thanks to both of you for the effort that you're taking to improve AudioMuse-AI.

ravenscroftj:
Thanks @NeptuneHub  - I think this PR that @HeroBrine1st  is working on will improve my initial work by making the NVidia docker container a lot smaller and quicker to download. That should benefit me too! 

I'm happy to support and test this work when it's ready and happy to collaborate on a shared vision :)

HeroBrine1st:
So what it needed from `devel` is cuda-nvvm package. I didn't look to logs hard enough yesterday but now it fully works (though I didn't run clustering yet).

What should be done so that it increases the speed (I'm focused on analysis task btw):

- Make analysis task a pipeline where each part is independent, works in parallel and as such does make GPU idle (song downloading, I'm looking at you!)
- Reuse loaded models as much as possible. Make inference of all models parallel, probably each in own task, and with workers that can filter tasks we will then have workers dedicated for one model, improving inference speed due to GPU memory caching (this is only if you use many GPUs, one GPU will still suffer from cache misses)
- Split the CPU worker from GPU worker (there is a task that simply picks tasks - that shouldn't be done by a worker that has GPU access)

This all will benefit CPU too due to simultaneous multithreading. Most of those tasks are memory bound, SMT is very good here.

Additional optimisation: even if there is only one worker, it can get tasks of the same type in bulk. Say we have 1000 tracks and 7 models. Instead of switching models for each track it is possible to do the same model in pipeline for 1000 tracks. You'll get only 1/7 of result, but the speedup will be very huge. Then you do the same for next model.

This means that database should be able to store incomplete data. That's not a problem.

HeroBrine1st:
cuda-compiler adds 260 MB on top cuda-nvvm's 40 MB but resolves warning about falling back to driver compiler. I think it somewhat increases performance but actually if one model is loaded once for multiple tracks it won't matter.

For me, this change only slows down analysis by two or even three times, so I'll go back to CPU image. We discussed possible optimisations, so I think we'll need to start with them first. But if your CPU doesn't have many cores or if you're on older memory this may speed things up. TL;DR Analysis is bound on I/O, network and memory bandwidth with GPU. With CPU, analysis is bound on network and either of CPU itself or its memory bandwidth.

Also I have found that all models combined take only half a GB on disk drive so this should take at most (pessimistic estimation) 2 GB, but it quickly ramps up to 8 GB. Idk why but it needs to be investigated (more memory to read means less performance (bandwidth is constant) and also more cache misses if the data is simply duplicated).

NeptuneHub:
@HeroBrine1st excuseme but I saw only now that you conferted the PR from draft to an actual one, I was thinking that you would first finish other task.
If your you possible have a check a resolve the conflict on Dockerfile? your new version is become pretty complex and I would like to avoid error (I mainly added a couple of new  library on the one on `main`)

HeroBrine1st:
That was too early. I forgot to set this back to draft. The HEAD of this PR has typo, and I have fixed that only locally (and still building to test the building)

Also I was in the middle of writing comment about the fix for #48 only in nvidia image which likely is not needed :-)

The typo is [here](https://github.com/NeptuneHub/AudioMuse-AI/pull/46/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R100), it needs to be `deployment`

NeptuneHub:
So I removed the typo in another commit, please let me know if something else need to be changed :)

HeroBrine1st:
Yes, without typo it builds at least without build args (checked on rootless docker). UPD: also builds with `BASE_IMAGE=nvidia/cuda:12.8.1-cudnn-runtime-ubuntu22.04`.

On fix - I found this in range-diff:

```
    +-# --- START FIX: Generate and set a proper UTF-8 locale ---
    +-# This prevents encoding errors with non-ASCII characters in track titles
    +-RUN apt-get install -y --no-install-recommends locales && \
    +-    sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
    +-    dpkg-reconfigure --frontend=noninteractive locales
    +-ENV LANG en_US.UTF-8
    +-ENV LANGUAGE en_US:en
    +-ENV LC_ALL en_US.UTF-8
    +-# --- END FIX ---
```

It was only in nvidia dockerfile and as I was unifying them I checked if it is nvidia-specific and it was not needed at all, so I removed that

NeptuneHub:
That part was for the issue  #48 but at the end the main probem was how he set the database, so I'm not pretty sure it was needed.

Now I'm looking that the `-essentia` build just failed (it used Dockerfile.essentia):

```
#17 [auth] neptunehub/audiomuse-ai:pull,push token for ghcr.io
#17 DONE 0.0s

#18 exporting to image
#18 exporting layers done
#18 exporting manifest sha256:1d926c57f3c620de28a926f2fd0e37e7ac7449794e8ef044fc9b8d6197c2daad done
#18 exporting config sha256:c2517640669de6fe803cf77eef2ab02a670bd7f1c7e76483e6acb9accf0f345d done
#18 exporting attestation manifest sha256:3c058fddae7bb341a813e0fa5cb0fdadf78900c06c47b9043409347f6c72b875 done
#18 exporting manifest list sha256:b415badc09f35488eb01fd5e2430b9d912b840d21e32fcce4a13485cfafd248a done
#18 pushing layers
#18 pushing layers 125.4s done
#18 ERROR: failed to push ghcr.io/neptunehub/audiomuse-ai:latest-essentia: unknown: blob unknown to registry
------
 > exporting to image:
------
ERROR: failed to build: failed to solve: failed to push ghcr.io/neptunehub/audiomuse-ai:latest-essentia: unknown: blob unknown to registry
Error: buildx failed with: ERROR: failed to build: failed to solve: failed to push ghcr.io/neptunehub/audiomuse-ai:latest-essentia: unknown: blob unknown to registry
```

The standard INTEL/ARM and the NVIDIA are still running

HeroBrine1st:
So the run in questrion is https://github.com/NeptuneHub/AudioMuse-AI/actions/runs/17469694061/job/49614616922. Precisely, it is this step:

https://github.com/NeptuneHub/AudioMuse-AI/blob/46043fe45b0a5bebe6e7ea3ffc589d933bf808f6/.github/workflows/build-intel-essentia.yml#L208-L216

This wasn't touched with this PR and essentia has a separate dockerfile, are you sure it is caused by this change?

NeptuneHub:
you're right, essentia have is own Dockerfile.essentia and even is own worfklow, so it as failed for "some other reason". I'm restarted just in case was failed "by chance".

Intel/Arm and Nvidia docker instead are compiled with success.. need to be tested just in case. Can you give a try to [latest-nvidia](https://github.com/users/NeptuneHub/packages/container/audiomuse-ai/506302572?tag=latest-nvidia) ? 

I'll give a try to [latest](https://github.com/NeptuneHub/AudioMuse-AI/pkgs/container/audiomuse-ai/506292620?tag=latest) on Intel and Arm


Edit `:latest-essentia` just with a re-start compiled, so was just a case :)

HeroBrine1st:
Yes, it successfully allocates memory on GPU and analyses songs (`SUCCESSFULLY ANALYZED` in logs). Haven't done any testing beyond that; I'm returning back to CPU images

NeptuneHub:
Great thanks. I tested the Intel and Arm image and it’s working fine. Looking forward for when you will have more time for further contribution :)

PR #47, opened 2025-08-17
fix: monkeypatch np.float_ to make clustering work again

ravenscroftj:
When running K-Means clustering, I was getting an AttributeError every clustering iteration:

```
AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.
```

I was able to trace this to the clustering module which is checking the type of a value from a list of types including `np.float_`, `np.float64` and so on. I did not do any kind of analysis over why the module checks for the old deprecated `np.float_` type and instead, to maintain backwards compatibility, I opted to monkeypatch the `np.float_`, setting it to `np.float64` if it does not exist.

This has now fixed clustering behaviour for me.

Tested and validated in `audiomuse-ai:latest-nvidia`

### DIFF
--- a/tasks/clustering.py
+++ b/tasks/clustering.py
@@ -42,6 +42,12 @@ from .clustering_helper import (
     _perform_single_clustering_iteration
 )
 
+# we want to maintain np.float_ for backwards compatibility but it was removed in numpy 2.0
+# the check below in sanitize_for_json causes an AttributeError that crashes the clustering algo
+# since it tries to access np.float_ so we monkeypatch np.float_ to point to np.float64
+if not np.__dict__.get('float_'):
+    np.float_ = np.float64
+
 logger = logging.getLogger(__name__)
 
 def _sanitize_for_json(obj):


HeroBrine1st:
It can be simply removed from relevant tuple where it is accessed:

https://github.com/NeptuneHub/AudioMuse-AI/blob/9a4fc1aa679def4dcbf15271b97fbca49690f48b/tasks/clustering.py#L61

Your change makes it ` elif isinstance(obj, (np.float64, np.float16, np.float32, np.float64)): `, which is effectively the same

ravenscroftj:
You're correct assuming there is no need to support np.float_ any more. I haven't tested across the other builds to see if there is still some older versions of numpy being used. If it's possible to force everyone onto numpy>=2.0 then I agree it would be better just to remove it from the tuple

NeptuneHub:
I had the [essentia version](https://github.com/NeptuneHub/AudioMuse-AI/blob/main/Dockerfile.essentia) that is locked to numpy==1.26.4.

I moved from essentia to librosa exactly for his old requirements and for the fact that I think is not still mantained, but I keep the essentia version for "legacy support".

@ravenscroftj if you can confirm that the code in the PR don't impact on who use older version, and only fix who use new, I'll proceed with merging it.

HeroBrine1st:
> I had the [essentia version](https://github.com/NeptuneHub/AudioMuse-AI/blob/main/Dockerfile.essentia) that is locked to numpy==1.26.4.

Didn't see that :-) Although I'd rather freeze the branch for essentia (or backport changes, or patch in Dockerfile), so that it is not broken by changes like this

Anyway this PR should not impact anything if `np.float_` is already present, so essentia should be good

NeptuneHub:
Ok so, if is safe as it is I'm proceeding with the patch (then I don't know still when I'll keep the essentia version)

HeroBrine1st:
https://numpy.org/doc/1.26/reference/arrays.scalars.html#numpy.double also see the aliases. np.float64 looks the same as np.float_ on numpy 1.26 already (on 64-bit systems though)

PR #121, opened 2025-10-20
Add Emby compatibility checks and duplicate Jellyfin API calls for Emby

moutasem1989:
I duplicated Jellyfin Server API Calls for Emby Servers. I copmared the used API endpoints to Emby Documentations and I added comments which is still compatble and which is not.
I have not tested it yet though.


NeptuneHub:
I'll wiat for your confirmation that the test work before proceeding with additional test on my side + merge.

I need to be sure that this don't create conflict with other mediaserver but as far as I see with a fast code review seems you're going in the correct direction. Thanks!

moutasem1989:
So I did check it on my machine with around 2k audio files and it managed to see all of them. I was able to use all functions without errors including creating playlists.
Please go over it because Python is really not my thing

NeptuneHub:
I merged but I'm doing some last test before build and release.
Can you share with which version of emby did you tested it meanwhile? Thanks!

PR #142, opened 2025-10-27
Reverse proxy support

### LINKED ISSUE #52: [FEATURE]
This is not exactly a major issue that needs to be addressed immediately, since my setup mostly works fine already but, I'll explain:

I have a server at home that runs Jellyfin, and unfortunately I don't have a dedicated static ip address at my home. I have to resort to using a VPS with a static ip to expose my server to the internet. From what it looks like, AudioMuse does not support path-based reverse proxying as opposed to using a subdomain, I could do a subdomain reverse proxy, but I prefer to keep all Jellyfin related things to a single subdomain.

For now, having Jellyfin directly communicate locally on my server works, but I can't access the frontend that AudioMuse provides because of the missing path-based reverse proxy support.

It would be appreciated if at some point later on, the flask app could support reverse proxies (both based on a subdomain and path-based)

mschaaf:
* generate URLs in templates dynamically ( :heavy_check_mark: )
* generate API URLs in javascript code dynamically ( :heavy_multiplication_x: )
* generate navigation URLs dynamically ( :heavy_check_mark: )
* add flag for explicitly enabling the feature (disable by default) ( :heavy_check_mark: )

Should fix #52


mschaaf:
@NeptuneHub the MR is ready IMO. Please check again. Thanks.

NeptuneHub:
Ok I tested the main functionality manually with both `ENABLE_PROXY_FIX` at `True` and `False` and I didn't see any regression. Off course don't having the proxy configured I can't test that it "do his job" but don't having regression is still a good test!

I'm going to merge this looking forward for the rest of the change needed to have the proxy fully work!

Really thanks for your support! 


Also run the smoke test in test/ just in case and all passed:
```
test.py::test_sonic_fingerprint_and_playlist [TIMING] Sonic Fingerprint test completed in 1.09 seconds
PASSED
test.py::test_song_alchemy_and_playlist [TIMING] Song Alchemy test completed in 0.39 seconds
PASSED
test.py::test_map_visualization [TIMING] Map Visualization test completed in 0.17 seconds
PASSED
test.py::test_annoy_similarity_and_playlist [TIMING] Similar Song test completed in 0.08 seconds
PASSED
test.py::test_song_path_and_playlist [TIMING] Song Path test completed in 0.59 seconds
PASSED
test.py::test_clustering_smoke_flow[kmeans-True-199] [RESULT] Algorithm=kmeans | BestScore=20.74596298608747 | PlaylistsCreated=8 | Time=58.74s
PASSED
```

PR #143, opened 2025-10-27
Reverse proxy support javascript 52

### LINKED ISSUE #52: [FEATURE]

mschaaf:
should fix #52

### DIFF
--- a/static/collection_script.js
+++ b/static/collection_script.js
@@ -160,7 +160,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
     async function checkActiveTasks() {
         try {
-            const response = await fetch('/api/active_tasks');
+            const response = await fetch(getActiveTasksEndpointUrl);
             const activeTask = await response.json();
 
             if (activeTask && activeTask.task_id) {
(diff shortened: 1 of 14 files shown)


mschaaf:
@NeptuneHub this is the last MR for the reverse proxy fix. For now I don't see any missing parts. Please check.

NeptuneHub:
Ok I tested it and I didn't noticed regression when the proxy is not present so I'm going to merge it.

longnguyen2004:
Hello, seems like you forgot to also update the setup related endpoints
https://github.com/NeptuneHub/AudioMuse-AI/blob/cd05e8596648b55e12e2d4e9391ac5b496912979/static/setup.js#L329

Since I don't have much experience with Flask, can you have a look? @mschaaf

NeptuneHub:
New release fix proxy, this time in a centralized way so hopefully it shouldn't come back:
- http://github.com/NeptuneHub/AudioMuse-AI/releases/tag/v2.3.1

Another user just tested it and seems working. If you still have problem with the new release please raise a new and dedicatred issue. thanks.

PR #153, opened 2025-10-29
Use .env for handling configurations variables

### LINKED ISSUE #111: [FEATURE] Better config handling
There are many duplicated config options between the 2 audiomuse-ai containers, and it might start behaving weird when the config isn't the same on both containers. To make matters worse, it is quite easy to miss when things aren't the same.

Most if not all the configs should be on one of the containers, and the other container should obtain config from the one that has them.

Shared config file mounted into both containers?

warreth:
fixes #111

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,8 @@ Choose the appropriate file based on your media server setup.
     ```bash
     cp .env.example .env
     ```
+    you can find the example here: [.env.example](.env.example)
+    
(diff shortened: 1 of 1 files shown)


NeptuneHub:
Great! Can I merge or do you still need to change something?

warreth:
Now you can merge i think

PR #159, opened 2025-10-31
Add reverse proxy setup as container

mschaaf:
* add all dependencies for a working audiomuse-ai setup with mediaserver


test/docker-compose.yaml
NeptuneHub: Can we add this change on a different file like `docker-compose-proxy.yaml` ? 

Most of the user will not need it, so I'll not overcomplicated their setup.

NeptuneHub (approved):

PR #196, opened 2025-11-20
Fix/openrouter rate limit and dmr

rendyhd:
A fix for:
- function now includes a retry mechanism with exponential backoff for handling rate limits (429 errors) and empty responses.
- included the DMR comparability


rendyhd:
Ow, this is on top of #195, sorry for the mess

NeptuneHub:
Is this still for clustering on GPU? can you add this change in the #195 so when finished I can download all in one and do only "one round of test" ? thanks.

rendyhd:
Included with #195

Issue #199, opened 2025-11-23
[BUG] 'tuple' object has no attribute 'replace'


marcchacon:
**Describe the bug**
The analysis checks the songs but does not save them on to the database. I think it's due to how Emby's API returns the data. In further inspection, there is a critical error on the logs: `[CRITICAL]-[22-11-2025 23-52-24]-Album analysis 27760 failed: 'tuple' object has no attribute 'replace'`

**To Reproduce**

0. the .env has the default values, except for the Emby login. It detects successfully the albums. (No connection error)
1. "Number of Recent Albums" set to 10
2. Click on Start Analysis
3. After the analysis, the status shows "task_state": "SUCCESS"
4. On Database Cleaning, the message is "Total tracks in database: 0"
5. When checking logs, the error is there, at least one from each song.

**Expected behavior**
It should detect the tracks and the value be at least 1.

**Screenshots**
<img width="960" height="555" alt="Image" src="https://github.com/user-attachments/assets/a8ba1fb5-0c79-424c-b1aa-5093be9b515c" />

<img width="910" height="280" alt="Image" src="https://github.com/user-attachments/assets/a35d989d-0395-4353-8515-01421a64e0ed" />

**Environment (please complete the following information):**
 - OS: Ubuntu 24.04.3 LTS x86_64
 - Deployment: Docker Compose
 - AudioMuse-AI Version: Version v0.7.11-beta
 - Emby Version: 4.9.1.90
 - Browser: Any (Last tried with Firefox 145.0.1 (64-bit) on Windows 10 and)
 - CPU: Intel i7-4500U (4) @ 3.000GHz (has AVX, but only 2 cores)
 - RAM: 2x4Gb DDR3 (1600 MT/s)
 - Disk: 1TB SATA SSD

**Additional context**
Looking at the logs, I found that the error that causes this is: 
```
audiomuse-ai-worker-instance  | [INFO]-[22-11-2025 23-52-24]-SUCCESSFULLY ANALYZED 'ISSEN by ('VESPERBELL', '27759')' (ID: 27758):
audiomuse-ai-worker-instance  | [INFO]-[22-11-2025 23-52-24]-  - Tempo: 170.45, Energy: 0.3843, Key: C# minor
[...]

PR #200, opened 2025-11-23
[FIX BUG] 'tuple' object has no attribute 'replace'

moutasem1989:
- `_select_best_artist(item, title)` returns `(artist_name, artist_id)` as a tuple.
- Old code assigned the entire tuple to `item['AlbumArtist']`.
- Downstream code expects `item['AlbumArtist']` (passed as `author`) to be a string so it can call `.replace()`.
- The patch unpacks the tuple and stores only the string in `AlbumArtist` and the `ID` separately in `ArtistId`.
- This prevents the `'tuple' object has no attribute 'replace'` error.

### DIFF
--- a/tasks/mediaserver_emby.py
+++ b/tasks/mediaserver_emby.py
@@ -442,7 +442,9 @@ def get_tracks_from_album(album_id, user_creds=None):
         # Apply artist field prioritization to each track
         for item in items:
             title = item.get('Name', 'Unknown')
-            item['AlbumArtist'] = _select_best_artist(item, title)
+            artist_name, artist_id = _select_best_artist(item, title)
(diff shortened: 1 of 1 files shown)


NeptuneHub:
This is the fix for #199 ?
Let me know when if finished, tested and ready to merge!
Thanks!

moutasem1989:
I think it is finished. I used `_select_best_artist(item, title)`wrong in one instance and the changes should fix it.

moutasem1989:
I tried it as was described in the bug and now the error is gone. It is safe to merge now

NeptuneHub:
Did you closed for error? Can I re-open and merge?

moutasem1989:
that was an error

Issue #254, opened 2025-12-27
[FEATURE] add port option for deployment in env file


NuxitBrain:
**Is your feature request related to a problem? Please describe.**
unfortunately, my port 8000 is already used by an other container. So i would like to change the mapping of the port with another

**Describe the solution you'd like**
simply create a PORT_VARIABLE in env file with will replace the 8000:8000 by PORT_VARIABLE:8000 in the docker-compose file.

**Describe alternatives you've considered**
N/A

**Additional context**
if your are ok with this solution, i can make a pull request with those edit. maybe the same should be done with other port (typicaly, people can have other instance of redis running ... )

NeptuneHub:
Hi ahd thanks for the feedback.
If you would like to raise a PR that add the variable in the .env.example file and then map it in all the different `docker-compose-*.yaml` flavour I'll merge it. Maybe you can do the same also for postgresql and redis port. thanks!

NeptuneHub:
Just merged, thanks for the help!

PR #255, opened 2025-12-27
add option to change default port with docker compose deployement

NuxitBrain:
in relation with the issue #254 

I add 3 variables in the .env.example :

- FRONTEND_PORT : port number for the flask app ( by default it's 8000) 
- WORKER_PORT : port number for the worker (by default it's 8029)
- REDIS_PORT : port number for the Redis crawler ( default it's 6379)

Normally all the yaml file have been changer and the older default port for the database have linked with variable POSTGRES_PORT ( not changed everywhere ...).

I also add a bit of doc for mentioning the existence of variable FRONTEND_PORT

I never used podman yet so i didn't change anything on this side ( don't want to make any mistake)

### DIFF
--- a/deployment/.env.example
+++ b/deployment/.env.example
@@ -48,7 +48,10 @@ POSTGRES_PASSWORD=audiomusepassword
 POSTGRES_DB=audiomusedb
 POSTGRES_PORT=5432
 POSTGRES_HOST=postgres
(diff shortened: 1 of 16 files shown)


NeptuneHub:
Thanks for the change, I'm going to merge. I never had the pleasure to use podman too, was from another contributor.

PR #311, opened 2026-02-04
Multi-provider & GUI Setup Flow

rendyhd:
## Summary

**Note!** Before testing please backup your production data. This will fully migrate the database on launch.

This is a foundational architecture change that adds multi-provider support to AudioMuse-AI, allowing users to connect multiple media servers simultaneously (e.g., Jellyfin + Navidrome + local files). It ships four tightly coupled features that were developed and tested together to avoid duplicate integration testing:

1. **Multi-provider architecture** — canonical `track_id` system with cross-provider deduplication
2. **LocalFiles provider** — scan and analyze local music directories without a media server
3. **Setup Wizard** (`/setup`) — guided web UI for adding/configuring providers, replacing manual `.env` editing for most settings
4. **Settings page** (`/settings`) — runtime configuration management for analysis, clustering, AI, and deployment settings

### Why one PR?

These features are deeply interdependent. The Setup Wizard is the primary interface for configuring multi-provider, LocalFiles is a new provider type exercising the multi-provider plumbing, and Settings removes the need for `.env` for most configuration. Testing any one of these in isolation would have meant re-testing the same integration paths twice. This has been running in production for ~2 weeks with Jellyfin, Navidrome, and LocalFiles simultaneously.

## Stats

- **105 commits**, 111 files changed, +18,837 / -4,551 lines
- **16 new files**, 15 deleted files (mostly replaced per-provider compose files with unified ones)

### Multi-Provider Architecture
- New DB tables: `track` (canonical identity), `provider_track` (links provider-specific `item_id` → `track_id`), `provider` (JSONB config registry)
- Cross-provider deduplication via `file_path_hash` (SHA-256) — same file from Jellyfin and Navidrome shares one analysis
- `resolve_track_id()` middleware in `app.py` — all APIs accept either legacy `item_id` or new `track_id`
- Automatic migration (`migration_track_id.py`, ~1,186 lines) — runs at startup, converts existing single-provider data to canonical schema
- Provider-aware playlist creation with target provider selector on all playlist pages
- Cross-provider `item_id` remapping for playlist creation (`remap_item_ids_for_provider`)

### LocalFiles Provider (`tasks/mediaserver_localfiles.py`)
- Scans local directories for audio files (configurable formats)
- Delta scan with mtime cache for fast re-scans
- M3U/M3U8 playlist creation with relative paths
- Album browsing, track metadata extraction via mutagen
- 620-line dedicated test suite

### Setup Wizard (`/setup`, `app_setup.py`)
[...]

PR #317, opened 2026-02-05
Improve playlist-title prompt variety

markjohnson303:
This PR updates the playlist title prompt used by AudioMuse. The previous prompt often produced repetitive titles that tended towards similar words and patterns (for example “neon”, “reverie”, “drive”).

The new prompt keeps the existing constraints (ASCII-only, single output, length bounds), but adds clearer guidance around originality and vibe, plus explicit “avoid trite phrasing” examples. In practice, this results in noticeably more varied, fun, and distinctive playlist names.

This prompt was tested with OpenAI gpt-5.2.


rendyhd:
I've ran a test across multiple models and while it performs great on GPT 5.2, it fails on many others - especially on smaller non-thinking ones often used in self-hosting. I've attached the results.

I agree with you, the names it gives with your prompt are a lot better, I really like much better.

To make it work I'm thinking of a couple of steps:
- Fine-tune the prompt to work on all thinking models.
- Add a toggle in settings, where the user can switch between the thinking and non-thinking prompt (and maybe even add a customize button). This is because a prompt aimed for a thinking model will remain problematic for non-thinking models, they generally handle contradicting instructions (pick 5, pick 1), negative instructions (do NOT), and longer prompts quite bad. Because of the model limitation I think a toggle is better than over complicating a system prompt and making it longer.

[ai_naming_OG_Prompt.html](https://github.com/user-attachments/files/25126353/ai_naming_OG_Prompt.html)
[ai_naming_PR_Prompt.html](https://github.com/user-attachments/files/25126355/ai_naming_PR_Prompt.html)

NeptuneHub:
I want to avoid additional button or parameter that overcomplicate the interface (and the avarage user don't use).

@rendyhd you are saying that the old prompt work in all those model but had more repetitive result, the new model instead work good in some model, and on other it totally fail? and fail mean that it put "the thinking" as a name or totally fail ?

My suggestion anyway is try to refine the prompt in order that work well everywhere.

Another point: if we have all this list of model that works with AI naming, if we check that they work good also on the Instant Chat functionality, let's add here (bottom of the page there is all the model supported):
https://github.com/NeptuneHub/AudioMuse-AI/blob/main/docs/PARAMETERS.md

Thansk.

rendyhd:
@NeptuneHub details are in the html including all responses, most failed because they gave the 5-10 options, gemini 3 pro gave a few empty responses after thinking.
image

markjohnson303:
@rendyhd thank you for digging into this so thoroughly! Do you have a tool you like for testing prompts across multiple models like that?

Definitely agree that as long as there is only one prompt, it should work reliably across all models. I can put some more work into that.

Another thought that I had was that perhaps it would be nice to allow users to personalize the prompt since there is so much variance between models. Clustered playlists seem like a headline feature to me, but I find them sort of frustrating because the names are so nondescript that it makes it hard to tell them apart.

rendyhd:
in #311 I've added a testing suite, in there you'll find ai_naming_test_config.yaml and test_ai_naming.py. That's the script I used to bulk test. It spits out the result in a results/ai_naming folder
It's set up for ollama and openrouter because it allows me to quickly go over many models - i didn't include the separate providers. In the config you can set the ollama ip (remember to enable network access in ollama), the openrouter api key, set the models, prompt, and hardcode a playlist

rendyhd:
This analysis on my full results might help you get in the right direction: https://claude.ai/public/artifacts/5d2d0634-97d9-4ec7-936c-e1c54049f27c

NeptuneHub:
Because I didn't see any update on this, I'm going to close.

PR #335, opened 2026-02-13
Improve search behavior on "playlist from similar song" feature

sfredo:
This PR streamlines track search behavior with the following changes:
1. Album names are now searchable
2. Song title, artist name and album names are combined in a single search field
3. Diacritic support (e.g., ó -> o, ã -> a)
4. Result ordering that prioritizes matches in song titles, then artist names, then album names (very open to suggestions)

I have only updated the "Playlist from Similar Song" feature so far, looking for an initial review before adjusting other sections.

No performance impact detected. Tested on a ~30K song library over a Raspberry Pi 5 without NVMe support

image

### DIFF
--- a/app_external.py
+++ b/app_external.py
@@ -121,20 +121,26 @@ def get_embedding_endpoint():
 @external_bp.route('/search', methods=['GET'])
 def search_tracks_endpoint():
     """
-    Provides autocomplete suggestions for tracks based on title and/or artist.
-    A query for either title or artist must be at least 3 characters long.
+    Provides autocomplete suggestions for tracks based on a unified search query
+    or legacy title/artist parameters.
+    A query must be at least 3 characters long.
     ---
     tags:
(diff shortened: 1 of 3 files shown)


sfredo:
Functionality extended to other features, PR is ready for review

NeptuneHub:
Hi and thanks for the contribution, very nice. For the future just a couple of suggestion:

- If not expressly agreeded avoid to modify legacy API, it can broke external API;
- The index for the search, I leave because is nice and I don't want to rework, but tested on 140k collection it move the search from 50ms to 25ms. Even on old architecture it wasn't never really slow. So I had the impression that we added line of code to mantain that wasn't really needed. For the future the less code for reach the scope is the best for an easy maintenance.

A part from this I two details really thanks for your contribution.

Note: you missed the music map search, but I fixed it easy.

Thanks.

PR #364, opened 2026-03-12
pin mistralai lib to below 2.0

zhol0777:
mistralai API changed in 2.0 release, leading to errors on worker start like "Error importing from app.py: cannot import name 'Mistral' from 'mistralai' (unknown location)"

### DIFF
--- a/requirements/common-noavx2.txt
+++ b/requirements/common-noavx2.txt
@@ -16,7 +16,7 @@ ftfy
 flasgger
 sqlglot
 google-genai==1.57.0
-mistralai
+mistralai==1.12.4
 umap-learn
 pydub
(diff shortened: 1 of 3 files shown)


zhol0777:
has been happening every time i attempt to build docker image regardless of release version of audiomuse-ai. pinning it to 1.12.4 (last release predating 2.0) gets it to work again
image

NeptuneHub:
Hi and thanks for sharing. Anyway the fix is already in `devel` branch. 

Important: the devel branch introduce a new machine learning model in place of CLAP, use DCLAP. so use devel only if you're ready to clean the database and start from a new fresh analysis.

Issue #377, opened 2026-03-21
[FEATURE] Default public visibility for auto-generated playlists


spcfox:
**Is your feature request related to a problem? Please describe.**
Automatically generated playlists in Navidrome are created as private, which makes them inaccessible to other users.

**Describe the solution you'd like**
I would like auto-generated playlists to be created as public by default. Alternatively, adding a configuration option to set the default visibility for auto-generated playlists would also be a great solution. Since Subsonic [supports public playlists](https://www.subsonic.org/pages/api.jsp#updatePlaylist), it seems logical for generated playlists to be visible to other users.

**Describe alternatives you've considered**
A global settings toggle for default playlist visibility would also address this issue.

NeptuneHub:
This is a really good idea, I’ll look the open subsonic api to check how they work. In this way the usage will be aligned to Jellyfin.

NeptuneHub:
Ok I implemented the fix for which the playlist will be public also in Navidrome to be aligned to what we already do with Jellyfin. The new `devel` image will be ready in around 30 minute. When ready please test and confirm that it's all ok.

Thanks for spotting this!

PR #378, opened 2026-03-21
Devel -> Main : create public playlist on navidrome - #377

NeptuneHub:
create public playlist on navidrome - #377

### DIFF
--- a/tasks/mediaserver_navidrome.py
+++ b/tasks/mediaserver_navidrome.py
@@ -393,11 +393,10 @@ def _create_playlist_batched(playlist_name, item_ids, user_creds=None):
     ids_for_creation = item_ids[:NAVIDROME_API_BATCH_SIZE]
     ids_to_add_later = item_ids[NAVIDROME_API_BATCH_SIZE:]
 
-    # Set public to true in line with OpenSubsonic / Navidrome createPlaylist expectations.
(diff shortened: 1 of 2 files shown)


spcfox:
I tried creating a "Playlist from Similar Songs" in this version, and it's still private.

NeptuneHub:
You download the devel image ?
ghcr.io/neptunehub/audiomuse-ai:devel

Because I put the public true tag on the update playlist and in the quick test that I did to me seems working.

If you downloaded the last devel image and you confirm that is not working means I need to doublecheck.

spcfox:
Yes, I’m using the `devel` image. I ran a few more experiments: if I create a large playlist (around 200 songs), it ends up public, but a small one (around 10 songs) remains private. As far as I understand, this is because `_create_playlist_batched` uses `updatePlaylist` for larger playlists.

NeptuneHub:
Nice spot, basically on Navidrome the playlist creation is made in more step:

1. the playlist creation add only a first part of song;
2. other song are then added with sequential update of playlist.

This because in the past we found user had issue when directly creating big playlist.

I added an additional update just to put public equal to true, this should fix the issue on small playlist that remain private.

Please double check also on your side that everything work (new devel image sohuld be ready in the next 30 minutes).

Thanks.

spcfox:
Now it works!

PR #406, opened 2026-04-09
N/A

ballfather:
-


NeptuneHub:
Hi and thanks for this PR. I want to say that the idea is interesting but let's share some high level requirements:

- Ok to have an Admin and an User role;
  - Admin role should see everything;
  - User role I'll suggest to hide all the administration functionality, but also the Analysis/Clustering => so basically it shouldn't be able to change the status of AudioMuse-AI itself or run high computational activity. User role could be helpful also to share a demo server
- In the administration menu, a submenu to admin user is needed. So ok if you use authentik to support it, but user without Authentik should be also be able to do everything in AudioMuse-Ai. Think that big part of the user have very simple deployment.
  - You should be able to create/delete/edit an user 
  - For each Music Server, the new user should also have the ability to still create playlist as "admin" OR setup his own music server user
  - An Admin can be able to edit/delete other admin user. It's important that at least an admin user remain
- It's important that all of this will work for all the music server so Jellyfin, Emby, Navidrome, Lightweight Music Server and Lyrion.
- it's important that all is documented with practical exemple. I would like to have at least example with Authentik deployed on Kubernetes and Traefik AND and example just with the use of AudioMuse-AI on a simple docker (this last one is like how most of the user use AudioMuse-AI).

A nice to have would be that user see only the library configured for his user. But this is BETTER be implemented in an FUTURE PR to not overcomplicate this.

To be ready the PR you need to share in human written word how do you addressed the above functionality and any additional functionality. You also share which test you did to proof that each of them worked on all the actual music server.


You can stay at high level in general, but I want to have low level details for everything that impact on:
- Database => like new table, new column;
- Music Server API => if you add additional call / modify the already present;

> IMPORTANT: Use of AI is accepted for coding, but with the above request I want to be sure that behind the AI there is a human that give correct direction and checked the output.

Excuseme if I ask a lot of details on this but for this kind of trasversal functionality is needed. One bug here can block everything.

ballfather:
-

ballfather:
-

NeptuneHub:
I'm happy to brainstorm with you this PR.

The requirements that I wrote is "how a multi user management" would like in its completeness. Say that I 100% agree to do in this PR just a small piece but having in mind the full picture.

So let's tailor back a smaller scope to be more manageble but still delivering value.

- Addressing the proxy part I think is the core of your PR.
- having 2 role, and admin and user, is also ok and very useful => I was thinking at this to put on a demo server.

Avoiding to extend to much this two point BUT giving the user to address this even WITHOUT Authentik (because I have Authentik, you have Authentik, most of user not) how we can proceed?
Would we want to start with two user, one per role, addressed by env var OR Authentik if you have, and then leave an administration page for future improvement?

Like in the env var / secret you place:

- ADMIN
- ADMIN_PASSWORD
- USER
- USER_PASSWORD

and if you use Authentik you can leave the password empty and just have the auth with Authentik?

Also remember that is not only about restricting the access to some html page, but also restricting the access to the API that they call (look the different functionality have a different `app*.py` page with all the route).

Really thanks for your effort on this.

EDIT: Off course let me know what do you think about this proposal, having just 1 user per role in the "local" implementation with ENV var I don't think is too complex and will give the benefit also to the user without authentik. I insist of this because for a DEMO server use cases I would like to have ADMIN and USER role without having to deploy Authentik.

rendyhd:
Hola! 
I also run Authentik (w/ Traefik). Implemented group-based auth via proxy on my home page recently. Happy to test when you're at that stage.

What's the use case? You have users interested in using the explore features? If so, do you want to let any user create playlists on your media provider viewable for everyone, or do you want playlists on user account level? What did you have in mind for AI usage / token costs?

NeptuneHub:
Here I think that the use case is simple like:
- you have an homelab;
- you deploy AudioMuse-AI and a music server in the homelab;
- You share the access to AudioMuse-AI AND the music server to your family&friends.

in this case administration functionality AND batch heavy functionality like Analysis/clustering shouldn’t be reacheable from everyone. And not only the html page but also the API.
Another use case that I have in mind is to share a Demo Server avoiding to share “distructive” functionality.

About chat functionality I think is totally an user functionality. If you don’t want to spend money you can just use a self hosted service.

The step of “each user with his own account” we said could be a next PR maybe linked to the possibility of the Ui configuration page that you’re are working on.

@ballfather if you like Jellify Discord server host an AudioMuse-AI channel here: https://discord.gg/yf8fBatktn

NeptuneHub:
Just to share that the Wizard Ui is merged (maybe some minor fix can come but nothing of impacting), so you can update your branch and complete this PR if you want. Thanks.

PR #410, opened 2026-04-11
Feature/provider migration tool

rendyhd:
This is the Provider Migration Tool based on my lessons learned of multi-provider, ported a lot of the matching logic over.
It's located in the admin menu.

I've successfully tested it on my production library Jellyfin -> Navidrome and Navidrome -> Jellyfin. 
I still have to spin up an Emby and Lyrion server to fully validate. Current implementation is based on my lessons learned.


NeptuneHub:
Thanks @rendyhd rornyour effort.
Can you describe in few words how to test it and at an high level how this work?

Before testing I’ll wait the implementation complete on Lyrion and Emby so I’ll do only one round of test.

Please on Emby and Lyrion pay double attention in testing because I’m also less expert on them and I can miss something. Especially Emby I never used, it was a contribution from another user.

rendyhd:
Yes, might be a bit before I have those servers up again.
From API perspective Jelly and Emby don't differ much, they're build on the same foundation.

You can test by going through the steps. I've done it with my production, turned off real path in navi, and added files on one server that weren't on the other.

It's a 6 step approach.
1- User agrees they made a backup
2- Connect to second server, it tested conenction and does initial file check on 100 tracks. If you e.g. have Navidrome not real path it gives a warning and suggests you turn that on and run it again.
3- It tries to auto-match, based on path and meta data. (you can click on any of these categories to filter in step 4)
4- You can check the matches and manually match anything that's not been matched by searching on the second server.
5- If you have matched everything (or deleted items you don't have in both), you can do a dry run. This validates and generated a .csv for you to check
6a- Execute migration, here user has to confirm again.
6b- After execution there are 3 more steps described. Update .env and .yaml, restart server, and run analysis.

image

NeptuneHub:
IF i understood correctly this require that:
- 2 server online at the same time;
- the two server need to be able to reach each other => if I have one on my homelab, and the second on internet becuase "any reason", maybe they are not reacheable and don't work;
- if the second server have the same song in a different path, it doesnt work;

Am I correct?

Do you think, in any way, could be possible to do something like:
- User do a backup from Server 1 of type1
- User import the backup to Server 2 of type2
- AudioMuse-AI do the import but, because the user said that is an import and not a simple restore, it mark it as import
- you run a batch reconciliation task (just to don't have the need of touch the analysis task and introduce error)


Here how to do the reconciliation I need your experience on this two otpion:

1. On one hand you say I can scan all the album, all the song of the album, if the path of this song match on my import, I have the new song id and the reconciliation is done;
2. If going by album, and song, we check Artist, Title, Album and if present Album Artist? we match 90%? 50%? less? The one where we can do a perfect match could be just re-analyzed.

Option 2 on my side should (correct me if I'm wrong):
- Have near 100% accuracy => If a song have same title, artist and album, and maybe also album artist, is very difficult if it's a different song. We want to say 99% just to admit that some different but same metadata duplicate can happen?
- It should cover big part of the case, I mean if the song are the exact same, they should have the exact same data. Some no-match could happen on how the different mediaserver create album? We can test this and understand which metadata use for match and improve this, but if we reach > 90% of song match, for an havarage library of 50k song, you need to rescan only 5000song, it's something duable. Different story if you don't match half of the library.

What do you think about?
Have you any information in this sense?
What bring you to use the song path? 


Also this could open other scenario where you can start exchange this export with other friends that in this way maybe don't have to re analyze the full library  (but here I'm going far).

rendyhd:
2 server online at the same time;
    the two server need to be able to reach each other => if I have one on my homelab, and the second on internet becuase "any reason", maybe they are not reacheable and don't work;
    if the second server have the same song in a different path, it doesnt work;



- AudioMuse-AI needs to be able to reach the NEW provider server. So if you are currently on Jellyfin and want to go to Navidrome, it needs to be able to reach Navidrome. It will used the data from current Database and match to the new server. If AudioMuse-AI can't connect to the new provider it doesn't work.
- If the second server has the same song in a different path it will try to match based on metadata, then normalized metadata, if that doesn't work you have to do it manually. That's step 3 and 4.**

---

Do you think, in any way, could be possible to do something like:
    User do a backup from Server 1 of type1
    User import the backup to Server 2 of type2
    AudioMuse-AI do the import but, because the user said that is an import and not a simple restore, it mark it as import
    you run a batch reconciliation task (just to don't have the need of touch the analysis task and introduce error)



- A backup of what? Of the database from Jellyfin or Navidrome? It uses the API connection just like analysis. If you can't reach the server, you also wouldn't be able to do an analysis?
- If you already have a backup of the database from the other provider (so you have the music matched to the correct provider item_id, you can just restore the backup -> you don't need the migration. The migration is there to update the item_id from the old provider to the new provider. AudioMuse-AI uses the item_id in all other places.

---

Here how to do the reconciliation I need your experience on this two otpion:

    On one hand you say I can scan all the album, all the song of the album, if the path of this song match on my import, I have the new song id and the reconciliation is done;
    If going by album, and song, we check Artist, Title, Album and if present Album Artist? we match 90%? 50%? less? The one where we can do a perfect match could be just re-analyzed.

Option 2 on my side should (correct me if I'm wrong):

    Have near 100% accuracy => If a song have same title, artist and album, and maybe also album artist, is very difficult if it's a different song. We want to say 99% just to admit that some different but same metadata duplicate can happen?
    It should cover big part of the case, I mean if the song are the exact same, they should have the exact same data. Some no-match could happen on how the different mediaserver create album? We can test this and understand which metadata use for match and improve this, but if we reach > 90% of song match, for an havarage library of 50k song, you need to rescan only 5000song, it's something duable. Different story if you don't match half of the library.

What do you think about?
Have you any information in this sense?
What bring you to use the song path?

Also this could open other scenario where you can start exchange this export with other friends that in this way maybe don't have to re analyze the full library (but here I'm going far).



- It checks first path, tries to do that automatically with some smart behavior to account for different path prefixes (/mount /mnt /mnt/music etc), but it's best to have both providers mounted to the same folder -> when you switch from Jelly to Navi, you start a new docker and have it look at the same folder.
- if paths fail, it then tries the combination of album/artist/track.
  - It always does first on album level, and then matches the tracks in an album.
- it then tries to normalize them and tries to match again
- anything it can't do automatically you have to do manually. You can search in in the new provider and select the album, OR choose to delete it from the database.

- Song path because if you switch provider but use the same library it's the only data we have that's 100% the truth. Providers can make matching mistakes leading to different metadata, if you have multiple releases of an album providers often mistake those (depending on how well you manage your metadata).

- after the matching you get a report with every single match, here you can validate.
- then if all is good, you can migrate -> here it will overwrite the item_id from old to new provider

rendyhd:
Finished all tests, have successfully migrated to all 4 providers. After migration I've ran an analysis, and created playlists on the new provider.

Couple of edits:
- Added collision detection (multiple albums matched to the one manually)
- Added better support for multi-disc metadata matching
- Added hot-reload. Still advices to restart docker, but works without restarting.
- Added better instructions on top

rendyhd:
@NeptuneHub haven't found any improvements or bugs anymore. Tested all providers successfully, even without path (with large library and without good path it is a lot of manual effort to match, but i can't change that - providers handle meta data). With good path, matches 1000+ albums automatically without any issues.

After every migration tested analysis and creating playlists, all worked.

Instructions should be clearer now, it's above the steps.

Added more guardrails, you can't do manual duplicates anymore.

Chaphasilor:
@NeptuneHub so I don't really have much to add here. The migration tool seems to match by path first (that's also the most reliable, since the files probably don't change when migrating to a new provider), and then falls back to using just the metadata.  
Metadata-based matching should definitely be fuzzy (not sure if it is yet), but won't be perfect. The best remedy is probably a nice UI for manually matching, although that's still not feasible for really large libraries.

But, is a missing match really that big of an issue? If matching fails, all that needs to be done is re-run the analysis for those tracks, and remove the dead tracks from the library. So there isn't really a way to lose data, the matching is just to save some time. The most important thing when "migrating" is probably just to keep all AudioMuse settings intact, but those shouldn't change anyway.  

This migration to switch providers is certainly less fragile than trying to manage multiple servers at the same time (because that *would* require perfect matching or duplicated databases).  
Hope this helps :)

NeptuneHub:
So I still need to find the time to test but if the approach is first try to do path check, and then fall back to fuzzy match on album, title and artist I think it’s ok because you just implemented both the approach.

About the test connection and the automatic switch to the new provider I want to finish the setup wizard (that I have in devel branch now) and then we will need to align the two functionality. Initially I wasn’t thinking you was implementing an hot swap. Anyway I don’t think so much rework will be needed.

rendyhd:
Just to clarify a couple of items:
### What does this branch do?

AudioMuse-AI stores analysis results (embeddings, mood, energy, etc.) linked to each track's **item_id** from your media server (Jellyfin, Navidrome, Emby, Lyrion). If you switch from one media server to another, all those IDs change — even though the actual music files are the same. Without this tool, you would have to re-analyze your entire library from scratch (which can take days for large libraries).

This branch adds a **6-step wizard** (in the Admin menu) that matches your old tracks to the new server's tracks, then rewrites the IDs in the database so all your existing analysis, embeddings, and playlists are preserved.

### What data gets overwritten?

The migration **replaces the item_id** (the track identifier) across every table:

- **score** — the main track table (also updates `file_path`, `title`, `author`, `album`, `year` to the new provider's values)
- **embedding, clap_embedding, mulan_embedding** — audio analysis vectors
- **playlist** — all playlist references
- **voyager_index_data / map_projection_data** — the similarity search indexes (ID mappings rewritten, actual vectors kept)
- **artist_index_data, artist_component_projection, artist_mapping** — these are **deleted** (they rebuild automatically on next use)

Tracks that have **no match** on the new server are **deleted** from the database.

### How does matching work? (Path → Metadata → Normalized Metadata)

The matcher uses **4 tiers, tried in order — first match wins**:

**Tier 1 — Normalized Path:** Both providers point at the same music files. Strip common mount prefixes (`/media/music/`, `/mnt/`, `/data/`, etc.), lowercase, compare. This is the most reliable because the file path is the actual truth — it's the same file.

**Tier 2 — Path Tail:** Compare only the last 3 folder/file parts (e.g. `Artist/Album/01 Song.flac`). Handles cases where mount points differ completely but folder structure is the same.

**Tier 3 — Exact Metadata:** Lowercase `(title + artist + album)` must match exactly. No tolerance for differences.

**Tier 4 — Normalized Metadata:** Same as Tier 3, but first strips noise like `(Remastered)`, `[Explicit]`, `(feat. X)`, `(Live)`, `[Radio Edit]`, leading "The ", extra whitespace. Then compares.

**Why not pure fuzzy matching (like "80% similar = match")?**

Fuzzy matching (e.g. Levenshtein distance, "similarity score > 0.8") is dangerous for music libraries because:

- "Love" and "Loved" are 80% similar but completely different songs
- "Hotel California" by Eagles and "Hotel California (Live)" by Eagles would match wrong versions
- Different songs with similar names by the same artist on the same album (remixes, interludes) would create false positives
- With a 50k track library, even a 1% false-positive rate = 500 wrong matches, and those tracks silently get the wrong embeddings

The normalized approach is **deterministic and predictable**: it strips known noise patterns (remastered tags, featuring credits) but still requires an exact match after stripping. You know exactly what it does, and it will never match two genuinely different songs. Anything it can't match automatically, the user handles manually in Step 4.

**LAST STEP - Manual Matching:** Lastly anything that isn't automatically matched can be manual matched through a search function. Anything that you can't match you can click delete on. All orphans will be removed during migration

### How does hot-reload work?

After migration completes:

1. The new provider credentials are saved to the `app_settings` database table (key: `migration.active_provider`)
2. The executing process immediately reads that row and updates the `config` module's variables in memory (e.g. `config.JELLYFIN_URL`, `config.NAVIDROME_URL`, etc.)
3. A Redis pub/sub message is published on channel `provider-migrated`
4. All other processes (Flask app, RQ workers) have a background thread listening on that channel. When they receive the message, they also read from `app_settings` and update their own `config` in memory.

Result: all processes switch to the new provider without restarting the container. (Restart is still recommended but not strictly required.)

### What needs to happen

1. **Coordinate with the Setup Wizard (PR #413):** Youmentioned wanting to align the hot-reload / provider-switching between the two features. The setup wizard restarts the container to apply config; the migration tool uses hot-reload via Redis pub/sub. These should use a consistent approach. **Let me know if I should remove to hot reload already**
2. **Reuse the connection test in the Setup Wizard:** The `tasks/provider_probe.py` file has a `test_connection()` function that can be wired into the setup wizard (integration instructions were already commented on PR #413).
3. **Update `.env` / `docker-compose.yaml`:** After migration, users need to update their environment variables to point to the new provider so the config survives container restarts. The migration wizard handles this in Step 6 instructions, but with the Setup Wizard storing config in DB, this should become automatic. **I propose that I'll wire that up once the setup wizard is done and remove the instructions.**
4. **Run analysis on unmatched tracks:** Any tracks that couldn't be matched just need to be re-analyzed, not the whole library. The migration preserves everything it can match.

rendyhd:
### Code review

Found 2 issues:

1. **XSS via `innerHTML` with unescaped external data.** `showError()`, `showWarn()`, and multiple rendering blocks inject media-server API responses and DB-sourced values (album names, artist names, error messages) directly into `innerHTML` without HTML escaping. An `esc()` helper exists at line 958 but is only used inside the collision-details table. All other insertion points — unmatched album list (line 671), candidate album list (line 830), connection-test warnings (line 600), error/warning banners (lines 519/525) — pass raw strings. A compromised or misconfigured media server could inject script tags via album or artist metadata.

https://github.com/NeptuneHub/AudioMuse-AI/blob/92bb8ab18b6981aa439a29bf2937e09f01aa9d60/templates/provider_migration.html#L518-L526

https://github.com/NeptuneHub/AudioMuse-AI/blob/92bb8ab18b6981aa439a29bf2937e09f01aa9d60/templates/provider_migration.html#L669-L674

https://github.com/NeptuneHub/AudioMuse-AI/blob/92bb8ab18b6981aa439a29bf2937e09f01aa9d60/templates/provider_migration.html#L828-L833

2. **Double pub/sub subscriber thread in `rq_worker.py`.** `subscribe_to_provider_migrated_channel()` has no idempotency guard — each call unconditionally spawns a new daemon thread (line 191 of `app_provider_migration.py`). When `rq_worker.py` does `from app import app`, the module-level block at `app.py:194` fires the first call. Then `rq_worker.py:33` calls it again explicitly. The comment at line 29 claims this is a "no-op second run," but it is not — two threads subscribe to the same Redis channel, causing `apply_provider_overrides_from_db()` to fire twice per migration event. This also violates the existing convention at `app.py:786`: "RQ workers import app.py but should NOT load indexes or start background threads" — the call at `app.py:194` runs unconditionally outside the `if not _is_worker:` guard used for all other background threads (lines 795, 874).

https://github.com/NeptuneHub/AudioMuse-AI/blob/92bb8ab18b6981aa439a29bf2937e09f01aa9d60/rq_worker.py#L28-L34

https://github.com/NeptuneHub/AudioMuse-AI/blob/92bb8ab18b6981aa439a29bf2937e09f01aa9d60/app.py#L192-L195

https://github.com/NeptuneHub/AudioMuse-AI/blob/92bb8ab18b6981aa439a29bf2937e09f01aa9d60/app.py#L786-L795

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

NeptuneHub:
As discussed in chat, please proceed with:

- resolve the conflict and update to the main branch;
- put the step with the step 4 with Manual album matching not mandatory
- add the possibility of a match only on TItle and Artist without album => put a disclaimer
- check how to solve the migration problem FROM and TO Navidrome

Thanks!

rendyhd:
Added: 
- Navidrome Real Path detection and rescan in step 3 (if you have real path off, it gives instructions, once you switch you can click rescan and will use path)
- Manual matching optional
- Option to match without album
- Removed hot reload

@NeptuneHub can you check with your test library again? My metadata is too clean, it just all works..

rendyhd:
Integration test failed because HuggingFace returned HTTP 503 Service Temporarily Unavailable when trying to download the RoBERTa tokenizer. Maybe rerun later?

tasks/provider_probe.py
NeptuneHub, line 1: ### MUSIC SERVER API CALL
In provider_probe.py you introduced API that call the music server (correct me if I miss some):

- fetch_all_tracks(provider_type, creds) — gets every track from the new provider
- get_album_tracks(provider_type, creds, album_id) — gets tracks for one album
- search_albums(provider_type, creds, query) — searches albums on the new provider
- test_connection(provider_type, creds) — validates connection and path format

Two of them are already covered in mediaserver*.py:
- fetch_all_tracks → already covered by existing get_all_songs() methods
- get_album_tracks → already covered by existing get_tracks_from_album(...)

For test_connection don't add additional call, just do like the Setup Wizard: it check a real provider connection by asking for top-played songs

The only missed one is:
- search_albums

This need to be implemented directly in mediaserver.py that then call ther specific music server API following the logic of the other API.

tasks/provider_migration_matcher.py
NeptuneHub, line 1: ### ARTIST + TITLE FALLBACK
There is an issue in provider_migration_matcher.py:

When allow_title_artist_only is enabled, the matcher builds a key from source metadata.
It was using the artist precedence:
- album_artist
- then artist
- then author

That is wrong for old rows, because source rows often have:

- author = real performer
- album_artist = Various Artists or generic unknown-album placeholder
So the current fix should instead use this source-side hierarchy:

- author
- artist
- album_artist
This ensures title + artist matching uses the real track artist from the old row, not a generic compilation/album artist.

tasks/provider_probe.py
NeptuneHub, line 1: ### ARTIST HIERARCHY
Please review provider_probe.py and make sure it normalizes remote provider metadata using the same artist-preference hierarchy that the provider-specific mediaserver_*.py logic expects.

Exact hierarchy I want checked:
Jellyfin / Emby
provider_probe should choose track-level artist in this order:

- ArtistItems[0].Name
- Artists[0]
- AlbumArtist
- fallback Unknown Artist

Then:

- artist should be that chosen track artist
- album_artist should still preserve the album-level AlbumArtist if available

Lyrion
provider_probe should choose track-level artist in this order:

- trackartist
- contributor
- artist
- albumartist
- band
- fallback Unknown Artist

Then:

- artist should be the chosen track artist
- album_artist should prefer albumartist or artist

Navidrome
This one already looks correct:

- artist
- albumArtist
- fallback Unknown Artist

Why this matters
provider_probe.py is the module that fetches metadata from the new provider and turns it into the unified shape consumed by the matcher. If its artist normalization differs from the provider-specific media logic, the metadata fallback can compare the wrong artist values.

NeptuneHub (changes requested):

tasks/provider_probe.py
rendyhd: how do you want pass target-provider creds into mediaserver.py (which reads from config)?
What you do for wizard, the _patch_config_for_test pattern isn't suitable for long and large processes right? Risky to introduce bugs.

NeptuneHub: Duplicate the API to music server is not a good idea.
Let’s try to define the use case and fine the best solution.
You have an user that want to migrate to another music server, so he want to change it.

Possibility 1 - the user first change the music server in the setup wizard, only then do the migration. In this way you totally skip the problem.

Possibility 2 - because you’re doing in step like a workflow the user test the music server, if it work he maybe apply. Apply could mean that you change the db value and you ALSO change the config parameter dynamically: which is the risk in a long procss? That you miss to change back ? But in this case you’re already making this change permanently 

Maybe you have other possibility in mind, I’m open to it.

Of course if you see issue on any of this possibility let’s discuss.

rendyhd: Both options have a blocker: the migration workflow needs to reach OLD and NEW providers in the same session, step 3 path refresh hits OLD (I added this after our discussion to let user rescan paths), the very next dry-run hits NEW, and step 4 manual matching is interleaved NEW calls. If either provider is reachable only via config, the other is not reachable. The long Jellyfin/Emby full-library pulls (100+ paginated calls per session) make _patch_config_for_test's mutate-and-restore window a minutes-long race surface (other Flask requests and RQ workers in the same process read the patched config)

The probe is not a real duplicate: 
- its return shape is unified (matcher is coded against it) whereas mediaserver_*.py has different flavours
- it includes track_number and disc_number, mediaserver doesn't
- its pagination is explicit (500/page + bounded timeout, whereas mediaserver's get_all_songs uses a single Recursive=true call that times out on 50k+ libraries 
- test_connection adds path-format classification which get_top_played_songs doesn't (matters for step 2 Warning, step 3 dry-run gate & path refresh)
- it takes creds from dict arg (avoiding the problem above)

If you use probe.test_connection for the Setup Wizard instead of _patch_config_for_test:
- no more duplication
- takes creds explicitly -> no _patch_config_for_test needed
- samples 100 tracks -> path-format classification possible
- returns warnings array -> Navidrome Real Path hint available during initial setup, not just during migration

NeptuneHub: So you highlight two point:

1. need of have a credential as input;
2. different logic of migration API vs the AS-is due to working per song instead of album, and additioanl parameters

About the **credential**, can we just add a NON MANDATORY `cred` parameters default to none, like other api already done? A minimal change that instead of using config.py use the one from the input ONLY when it's not none.
In this way legacy logic shouldn't have issue, a new api can just pass the credential.

About the **different logic** is there a way to reuse the new one?
Instead of call all the song item paginated, can we reuse the analysis approch of get list of album and then get song for album?
About the extra field, can we just read them in the actual API? 

Then if there is a totally new API never used, ok to add it, but adding it following the logic of the meidaserver.py file.

Here we need to achive this:
1. Have the mediaserver call centralized in the same file with the same approch;
2. Try to reuse what we have, or do minimal adapatation, as much as possible.

The final goals is to avoid to increase the interconnection between AudioMuse-AI and the other music server. Otherwise an update of an existing music server or adding a new one, will raise the maintenance work.

rendyhd: Would all those additional changes not be out of scope of this PR? You want changes as small as possible.
If I understand:
- We park this PR
- Create new PR for restructuring mediaserver and others
- Restructure this PR

Before we park it, mind testing if the logic works on your library?

NeptuneHub: So the matching logic is very good, with your last fix we have:

- Only Metadata - Artist/Album/Title => 9530 matched out of 18925
- Only Metadata - Artist and Title only => 10.358 out of 18925
- Matched by path (used Navidrome path recovery) => 18882 matched out of 18925

This is very good because I used Cretive Common songs from different source and the quality of metadata on most of them was very poor. So if you think to the quality of metadata this result are very very good.

The last missing improvement that I need to merge this PR is this part related to the API.

rendyhd: Great, happy to see those results! Means the migration logic is done.

For the API part, as I shared above, I believe that should a separate PR. Looks like a large effort with potential regression.

NeptuneHub: If you think is easy to address with a different PR please proceed. But the additional PR I suppose should also include this change in order to test all together, so which are the advantages?

Edit: anyway for me don’t make any difference as soon as then can be tested all together.

NeptuneHub: I try to do a **NON PRODUCTION READY** patch for the API topic that  is this one:

[migration_20260416_120508.patch](https://github.com/user-attachments/files/26780429/migration_20260416_120508.patch)

I created the patch with this command on ubuntu:
```
git diff > migration_$(date +%Y%m%d_%H%M%S).patch
```

if you move on the pr-410 branch, on this commit:
```
Commit message: Merge branch 'main' into feature/provider-migration-tool
Commit hash: 314a912ff74fd33ea3ab386540855cd5022032b6
```

you can then apply with this:
```
git apply migration_20260416_120508.patch
```


The patch move all the api call on mediaserver*.py
This API was added from the probe file to the mediaserver:

* `search_albums` 
* `test_connection `

Instead this other was deleted to match already existent API:
* `fetch_all_tracks(provider_type, creds)`
* `get_album_tracks(provider_type, creds, album_id)`


I tested the migration from Navidrome to Lyrion, and all the use cases give me back the same exact number and result that I wrote in a test above. 

Only the test connection was tested also on Navidrome, Lyrion and Jellyfin.

Maybe you can start from this and test all the combination and fix any remaining issue? 

IMPORTANT: this is only an experiment of the morning. I thought that having the API in the expected format could help with the reworking but still need appropriate testing.

NeptuneHub:
Added to the PR branch my fix after doing the below migration test.
I don't have emby configured to test it too.

```
Navidrome -> Lyrion
- test connection: ok
- automatic matching / with album / proceed with medata only matching: ok
- automatic matching / without album / proceed with medata only matching: ok
- automatic matching / refresh source paths: ok (18883)
- manual album matchin (search): ok
- finalyze dry run / execute migration: queued / started: ok
- run full analysis: ok
- test music map / song path: ok


Lyrion -> Navidrome
- test connection: ok
- automatic matching / with album / proceed with medata only matching: ok
- automatic matching / without album / proceed with medata only matching: ok
- automatic matching / refresh source paths: N/A
- manual album matchin (search): ok
- finalyze dry run / execute migration: queued / started: ok
- run full analysis: wasn't needed ?
- test music map / song path: ok

Navidrome -> Jellyfin
- test connection: ok
- automatic matching / with album / proceed with medata only matching: 
	KO: Sanitization, fixed
	OK (302+1)
- automatic matching / without album / proceed with medata only matching: 
	OK (302+1+129)
- automatic matching / refresh source paths: N/A
- manual album matchin (search): ok
- finalyze dry run / execute migration: queued / started: ok
- run full analysis: run just with 1 track, ok
- test music map / song path: ok

Jellyfin -> Navidrome

- test connection: ok
- automatic matching / with album / proceed with medata only matching: ok
- automatic matching / without album / proceed with medata only matching: ok
- automatic matching / refresh source paths: N/A
- manual album matchin (search): ok
- finalyze dry run / execute migration: queued / started: ok
- run full analysis: run just with 1 track, ok
- test music map / song path: ok

Jellyfin -> Lyrion

- test connection: ok
- automatic matching / with album / proceed with medata only matching: ok
- automatic matching / without album / proceed with medata only matching: ok
- automatic matching / refresh source paths: N/A
- manual album matchin (search): ok
- finalyze dry run / execute migration: queued / started: ok
- run full analysis: run just with 1 track, ok
- test music map / song path: ok


Lyrion -> Jellyfin

- test connection: ok
- automatic matching / with album / proceed with medata only matching: ok
- automatic matching / without album / proceed with medata only matching: ok
- automatic matching / refresh source paths: N/A
- manual album matchin (search): ok
- finalyze dry run / execute migration: queued / started: ok
- run full analysis: run just with 1 track, ok
- test music map / song path: ok

```

NeptuneHub:
Also after all the round of migration I did a test on Jellyfin of all the other feature for regression checking.
The migration needed an autoamtic restart otherwise sonic fingerprint functionality don't update the configuration.

```
On Jellyfin:
Playlist from similar song + Playlist creation: ok
Artist similarity + Playlist creation: ok 
Song path + Playlist creation: ok 
Song Alchemy (2 mood in add) è Playist creation: ok
Text Search + pèlaylist creation: ok
Sonic Fingerprint: KO => don't update the credential, fixed with an automatic restaer


On Navidrome:
Playlist from similar song + Playlist creation: ok
Artist similarity + Playlist creation: ok 
Song path + Playlist creation: ok 
Song Alchemy (2 mood in add) è Playist creation: ok
Text Search + pèlaylist creation: ok
Sonic Fingerprint: OK

On lyrion:
Playlist from similar song + Playlist creation: ok
Artist similarity + Playlist creation: ok 
Song path + Playlist creation: ok 
Song Alchemy (2 mood in add) è Playist creation: ok
Text Search + pèlaylist creation: ok
Sonic Fingerprint: OK
```

NeptuneHub:
I think this is tested enough, I'm going to merge to keep all aligned with main branch.
If you find any issue on the emby side, please let me know and eventually raise a PR for fix. thanks.

PR #411, opened 2026-04-11
fix(ui): dark-mode restore warning and scrollable sidebar

rendyhd:
- **Backup page:** the `.restore-confirm-area` warning box used a hardcoded light-pink background with no dark-mode override, and the inline-styled `<p>` inside pinned its color to `var(--text-main)` (near-white in dark mode), producing unreadable light-on-light text. Added a `html.dark-mode` / `body.dark-mode` override that switches the box to a translucent red background and forces readable text (including an `!important` on the inline-styled `<p>`).
- **Sidebar:** `.sidebar` was `height: 100vh` with no `overflow`, so nav items past the viewport were clipped and unreachable. Added `overflow-y: auto` plus `overscroll-behavior: contain` to prevent scroll-chaining to the page body.

### DIFF
--- a/static/menu.css
+++ b/static/menu.css
@@ -9,6 +9,8 @@
     color: white;
     padding: 1.5rem;
     height: 100vh;
+    overflow-y: auto;
+    overscroll-behavior: contain;
     position: fixed;
     top: 0;
     left: 0;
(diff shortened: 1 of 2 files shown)


rendyhd:
First one:
From:
image

To:
image

Second one:
From (not scrollable):
image

To (scrollable):
image

NeptuneHub (approved):

PR #421, opened 2026-04-13
fix(docker): add .gitattributes to enforce LF for container files

rendyhd:
## Summary

  Fixes Docker containers failing to start with `bash\r: No such file or directory` on Windows-built images.

  ## Problem

  Commit cf50c81 introduced three new files that execute inside Linux containers: `docker-entrypoint.sh`,
  `restart_listener.py`, and `restart_manager.py`. On Windows machines with `core.autocrlf=true` (the default), Git
  checks out these files with CRLF line endings in the working tree. When `docker build` runs `COPY`, it copies the
  working tree as-is — so the container receives files with `\r` at the end of every line. The shell interprets the
  shebang as `#!/usr/bin/env bash\r`, which doesn't exist, and the container crash-loops.

  This was never an issue before because the project had no shell script entrypoint — all prior files were either Python
   (which tolerates CRLF) or config files that weren't executed directly.

  ## Fix

  Adds `.gitattributes` rules that force `eol=lf` for file types that run inside containers (`*.sh`, `*.py`, `*.conf`,
  `Dockerfile*`). This overrides `core.autocrlf` per-file, so Git always checks them out with Unix line endings
  regardless of the host OS.

  **Note:** Existing clones need to re-checkout affected files (or re-clone) for the rules to take effect.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,5 +4,4 @@
 *.sh text eol=lf
 *.py text eol=lf
 *.conf text eol=lf
-Dockerfile text eol=lf
-Dockerfile-noavx2 text eol=lf
+Dockerfile* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,8 @@
 *.pb filter=lfs diff=lfs merge=lfs -text
+
+# Force LF for files that run inside Linux containers
+*.sh text eol=lf
+*.py text eol=lf
+*.conf text eol=lf
+Dockerfile text eol=lf
+Dockerfile-noavx2 text eol=lf
--- a/Dockerfile
+++ b/Dockerfile
@@ -364,7 +364,11 @@ RUN set -eux; \
 COPY . /app
 COPY deployment/docker-entrypoint.sh /app/docker-entrypoint.sh
 COPY deployment/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
-RUN chmod +x /app/docker-entrypoint.sh
+# Strip CRLF from container files in case they were checked out on Windows
+# with core.autocrlf=true (git's initial-clone checkout may not honor
+# .gitattributes eol=lf, which would break the entrypoint shebang).
+RUN sed -i 's/\r$//' /app/docker-entrypoint.sh /etc/supervisor/conf.d/supervisord.conf \
+ && chmod +x /app/docker-entrypoint.sh
 RUN ls -l /etc/supervisor/conf.d && test -f /etc/supervisor/conf.d/supervisord.conf
 
 # ============================================================================
--- a/Dockerfile
+++ b/Dockerfile
@@ -364,11 +364,7 @@ RUN set -eux; \
 COPY . /app
 COPY deployment/docker-entrypoint.sh /app/docker-entrypoint.sh
 COPY deployment/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
-# Strip CRLF from container files in case they were checked out on Windows
-# with core.autocrlf=true (git's initial-clone checkout may not honor
-# .gitattributes eol=lf, which would break the entrypoint shebang).
-RUN sed -i 's/\r$//' /app/docker-entrypoint.sh /etc/supervisor/conf.d/supervisord.conf \
- && chmod +x /app/docker-entrypoint.sh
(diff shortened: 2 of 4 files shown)


.gitattributes
```diff
@@ -1 +1,8 @@
 *.pb filter=lfs diff=lfs merge=lfs -text
+
+# Force LF for files that run inside Linux containers
+*.sh text eol=lf
+*.py text eol=lf
+*.conf text eol=lf
+Dockerfile text eol=lf
+Dockerfile-noavx2 text eol=lf
```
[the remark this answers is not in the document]
NeptuneHub: @rendyhd can you apply this suggestion and look if in this way still work? Then I’ll merge.

NeptuneHub:
Ok, I'm going to merge.

Important: This only takes effect on files after Git normalizes them. After updating .gitattributes run:

```
git add --renormalize .
```

rendyhd:
I'm still testing, the --no-chache rebuilds take a while

NeptuneHub:
Ok, confirm when your test succed and I'll merge this pr. thanks.

rendyhd:
Final test done. Last commit fully fixed it.

  - --no-cache build from pristine fresh clone on Windows: succeeded (zero layer reuse)
  - Sed layer executed fresh: # 33 DONE 0.2s (no CACHED marker anywhere)
  - In-image files: both LF
  - Flask app: running cleanly, entrypoint fully executed, Voyager/map subsystems alive

  Every link in the chain proven end-to-end, from zero-cache build context containing CRLF bytes, through the Dockerfile's sed step, to a container whose shebang executes and whose app reaches steady state.

---

Why .gitattributes alone wasn't enough

  .gitattributes with *.sh text eol=lf is supposed to override autocrlf and force LF on checkout. But on git 2.52.0.windows.1, this doesn't work for git clone's initial checkout when autocrlf=true. We proved this directly:

  - Pristine git clone of the fix/gitattributes-lf-enforcement branch → docker-entrypoint.sh landed on disk as CRLF
  - file(1) confirmed: "Bourne-Again shell script, with CRLF line terminators"
  - Container built from that clone failed with the exact 'bash\r' error

So a brand-new Windows contributor cloning this repo and running docker compose up --build would still hit the original bug. .gitattributes is effective on subsequent per-file checkouts, but it doesn't fix the initial clone.

 git add --renormalize . didn't fix it

  What 97f3405 does

  Added to Dockerfile and Dockerfile-noavx2:

  RUN sed -i 's/\r$//' /app/docker-entrypoint.sh /etc/supervisor/conf.d/supervisord.conf \
   && chmod +x /app/docker-entrypoint.sh

  Strips \r from the files inside the image during build, so whatever git put into the build context (LF on Linux/Mac, CRLF on Windows) becomes LF before the image is finalized. The entrypoint shebang works regardless.

Dockerfile
NeptuneHub, line 1: I think that the cleanest approch is:
- Line ending test on pull request that fail in case of wrong end of line => I just added, so we will avoid to merge this error in the codebase;
- Keep the .gitattributes as you set;
- For windows user: Run `git add --renormalize . `  before compiling locally or push to the repository

I think that is not good having something that change the code at build time in the Dockerfile. Please remove.

Dockerfile-noavx2
NeptuneHub, line 1: Same comment of Dockerfile.

NeptuneHub (changes requested):

Dockerfile
rendyhd: I tried git add --renormalize .
It staged zero files, because the repo's index already stores LF, and the "clean" filter reading a CRLF working tree just produces LF, matching the existing blob. No diff to commit, nothing to push, nothing to change the clone behavior for future users.

rendyhd:
I've reverted the fix

Dockerfile
NeptuneHub: I think that the problem is only on this file `deployment/docker-entrypoint.sh`

if with the new .gitattributes you force git to recompute with this command (check that the path is correct):
```
# 1. Remove ONLY the problematic file from the Git index
git rm --cached deployment/docker-entrypoint.sh

# 2. Add it back (Git now forces it to LF because of the .gitattributes)
git add deployment/docker-entrypoint.sh
```

It should fix.

PR #422, opened 2026-04-14
Consolidate configuration import patterns

ballfather:
OK this is a very noisy patch BUT it's very simple in nature.

Don't think I need to explain the benefits of this patch.

The last patches from #413 [included](https://github.com/NeptuneHub/AudioMuse-AI/pull/413#pullrequestreview-4100966482) the beginnings of this effort. This patch concludes this effort.

Have I handchecked everything? Yes
Have I verified that this does not break logic in all cases? Yes
Have I actually tested the code? Not yet!! Doing so now.


ballfather:
Hello Gemini, your review is correct, but what is the point of all that code if AUDIO_LOAD_TIMEOUT ends up not being read by anything in this module? Why have this module check for a variable it doesnt use? Why not propose a patch to check all possible ENV variables that could ever exist across all namespaces?

PR #423, opened 2026-04-14
WIP: Fix `refresh_auth_state()`

ballfather:
- `auth_configured` and `bootstrap_auth_mode` were not being assigned values the exact same ways.
- We now assign the stripped version to `effective_audiomuse_user` and `effective_audiomuse_password`: if you're going to check for  the stripped version, might as-well directly use the stripped version as well (its indeed the proper thing to do in this context)

### DIFF
--- a/app.py
+++ b/app.py
@@ -124,22 +124,6 @@ def is_bootstrap_mode():
     return not config.AUTH_ENABLED or bootstrap_auth_mode
 
 
-def refresh_auth_state():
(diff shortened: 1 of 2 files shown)


ballfather:
OK Gemini interesting review! Are we sure we want to introduce dynamic ENV credentials? Is it even possible to reload ENVs for a docker without restarting it?

NeptuneHub:
I think is already good enough how the main branch is now: config parameter are static, to be changed the save button restart the container.

I think is enough because people shouldn't change parameter multiple time per days, so a restart of the container to start form a clean situation I don't think is so bad. Also I look it takes just some seconds with supervisord (maybe a bit more if you have big index to reload).

ballfather:
Needs more work for now. Will reopen once things are clearer.

PR #425, opened 2026-04-15
N/A

ballfather:
-


NeptuneHub:
Hi and thanks for this PR.

Can you also write the use case that bring the issue, so that I can do my test too?

Will be useful if you have a database backup that you can share (WHITOUT song in it, with user and password set to test) or any way to reproduce the bug. Thanks.

tasks/setup_manager.py
```diff
@@ -158,16 +158,26 @@ def _is_valid_auth_config(self, config_module):
...
         )
 
+        # TODO: Behaviour ambiguity
+        # Wizard's UI is designed in a way that allows you to complete the setup without entering media server details
+        # but this function pretends that setup is not complete until you do.
+        # Looks like either WizardUI is incorrect, or the specs for this function are incorrect.
+        # For now we will allow setup to be considered complete as long as auth config is valid, even if media server config is not.
+        #
+        # return server_valid and auth_valid
+        return auth_valid
```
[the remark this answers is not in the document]
ballfather: Indeed. But we need to clear design ambiguity first! This holds for now, or change Wizard behaviour as well!

ballfather:
-

NeptuneHub:
Get Gemini review just as suggestions, sometimes get to much worried :)

app.py
```diff
@@ -299,7 +261,7 @@ def login_page():
             pass
 
     auth_warning = None
-    if config.AUTH_ENABLED and (not AUDIOMUSE_USER or not AUDIOMUSE_PASSWORD):
+    if config.AUTH_ENABLED and (not effective_audiomuse_user or not effective_audiomuse_password):
```
ballfather: Yeah here we're using the `effective_audiomuse_xxx` instead of the raw!

app_setup.py
```diff
@@ -208,15 +208,13 @@ def setup_api():
...
 
-        was_bootstrap = is_bootstrap_mode()
         new_server_type = filtered_values.get('MEDIASERVER_TYPE', config.MEDIASERVER_TYPE)
         if new_server_type != config.MEDIASERVER_TYPE:
             obsolete_fields = config.MEDIASERVER_OBSOLETE_FIELDS_BY_TYPE.get(new_server_type, [])
             if obsolete_fields:
                 setup_manager.delete_config_values(obsolete_fields)
         setup_manager.save_config_values(filtered_values)
         config.refresh_config()
-        refresh_auth_state()
```
ballfather, line 219: -

NeptuneHub:
I need that you explain better what you want to achive with this PR to better understand it.

Just by analyzing the code, but maybe I'm wrong, we have:


### Current behavior in pr-425

- is_setup_complete() now returns true if auth config is valid, even when media-server vars are missing.
- /login redirects to /setup when setup is incomplete.
- require_setup_completion() now incorrectly redirects normal pages to /setup when setup is complete.

### My Intended initial behavior

- If mandatory variables are missing (Music server configuration, AudioMuse-AI login), show the wizard.
- Otherwise go to the normal landing page.
- If auth is enabled and AUDIOMUSE_USER/AUDIOMUSE_PASSWORD are configured, require login. But only if the mandatory variables are configured (so setup complete)

### Difference

1. pr-425 weakens the wizard check by only validating auth, not all required config.
2. It also has an inverted setup guard that can redirect users to /setup even after setup is complete.

So on the 1. I think is intended by you, but please explain te pro and cons of this change.
For the 2. probably is unintended ?

About the refresh_auth_state() shouldn't be needed anymore but off course test just in case :)

ballfather:
-

ballfather:
-

Issue #431, opened 2026-04-17
[BUG] v1.0.0+ can no longer be run as non-root user


waratahmidwifery:
**Describe the bug**
The latest update can no longer be run as a non-root user in docker.

**To Reproduce**
```
audiomuse:
  image: ghcr.io/neptunehub/audiomuse-ai:0.9.6
  container_name: audiomuse
  restart: 'no'
  user: 1132:1132
  env_file: /mnt/Apps/docker-data/navidrome/.env-audiomuse-ai
  environment:
    SERVICE_TYPE: "flask"
```

**Expected behavior**
Expect it to work with a non-root user (as was possible with 0.9.6 and earlier)

**Environment (please complete the following information):**
 - Deployment: Docker Compose
 - AudioMuse-AI Version: 1.0.1

**Additional context**
The logs of failed startup simply show:
```
root@server[/mnt/Apps/docker-setup]# docker logs audiomuse               
ENTRYPOINT: SERVICE_TYPE=flask
ln: failed to create symbolic link '/etc/localtime': Permission denied
```
Removing `user: 1132:1132` from the config lets it start again (though this is now running as root).

NeptuneHub:
Hi and thanks for the feedback. I added a fix in `devel`, the new image should be ready in around 30 minute.

Can you test and confirm that now works without issue?

Thansk.

waratahmidwifery:
Thanks for the prompt fix!
Confirming that the devel image resolves this 😄

PR #432, opened 2026-04-18
Fix timezone var to avoid root priviledge - #431

NeptuneHub:
Fix timezone var to avoid root priviledge - #431


mreid-tt:
Hey @NeptuneHub, I love the new v1.x release. I do note in the UI that all reported times on the dashboard seem to be in UTC rather than the TZ configured in my env. Is something wrong on my end or yours?

PR #435, opened 2026-04-18
Fix item_id collision when source and target IDs overlap (Emby <-> Lyrion)

rendyhd:
Postgres enforces PRIMARY KEY / UNIQUE row-by-row during UPDATE (they are not deferrable by default), so the single-pass UPDATE failed whenever a mapping's new_id equaled another row's current old_id — hit in practice on Lyrion -> Emby, where both providers use small integer IDs and "25" appeared as both an existing score.item_id and a target new_id for a different row.

Rewrite the score / playlist / embedding / clap_embedding / mulan_embedding UPDATEs as a two-pass per table: pass 1 stages every row at <prefix>||new_id (unique per new_id, cannot collide with any real id), pass 2 strips the prefix to land the final new_id. Earlier fix 5c80298 only addressed the orthogonal case of two old_ids mapping to the same new_id via the UNIQUE(new_id) guard on item_id_migration_map.

### DIFF
--- a/tasks/provider_migration_tasks.py
+++ b/tasks/provider_migration_tasks.py
@@ -48,6 +48,15 @@ _ADVISORY_LOCK_KEY = 7421536190082003
 # How long to wait for in-flight RQ jobs to drain before forcing migration.
 _DRAIN_TIMEOUT_SECONDS = 60
 
+# Intermediate prefix used during the two-pass item_id rewrite. Postgres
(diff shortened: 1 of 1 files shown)


NeptuneHub:
Before merging:
- give a check to gemini suggestion;
- do some regression test also in other music server: it shouldn't create impact, but just in case and confirm that it's ok.

Thanks.

Issue #107, opened 2025-10-15
[FEATURE] Better error messages in the web interface


felix920506:
**Is your feature request related to a problem? Please describe.**
When I first set this up I was getting confusing behavior (it tells me no new albums on first run which is weird) and only found out that the connection to my Jellyfin server from the worker was not working after checking the docker logs. 

The error should be reported in the web ui instead of silently failing

**Describe the solution you'd like**

Properly show errors on the web interface 

**Describe alternatives you've considered**

**Additional context**

NeptuneHub:
Thank you for your feedback! I agree that your suggestion could help improve the first-time setup experience. However, as the sole maintainer of this open-source project, it may take me some time to research and implement it due to other priorities.

If you’re interested, I’d really appreciate your contribution, you could explore, develop, and test improved error handling.

**Security hint:** From a security standpoint, it’s best not to display detailed error messages in the web interface, as they can reveal information useful to attackers (for example, whether a username or password is valid). Instead, use generic messages such as “invalid credentials” or “missing configuration” to avoid information disclosure.

I’m looking forward to any possible contributions, I’ve marked this as a good first issue.

rightsaidfred99:
Agreed. I got audiomuseai running but for the life of me can't figure out why I cant get it to talk to Navidrome server. A helpful message would be great for non techy people.

jackmcintyre:
> Agreed. I got audiomuseai running but for the life of me can't figure out why I cant get it to talk to Navidrome server. A helpful message would be great for non techy people.

I think this is bigger than error messaging, as there is also no logging or other indication that it failed (if your error is the same as mine). I created https://github.com/NeptuneHub/AudioMuse-AI/issues/330 to track separately.

aksPatel08:
Hi, I’d like to work on this issue. I’ll submit a PR shortly.

MichelJenishDev:
Hi , I'd would like to work on the issue. , I 'll submit a PR shortly. i  would do a proper documentatio about the features and things which can be impove to make the a good reach among the developers.

camicasa:
Hello! I worked on this issue and submitted a PR! Please check it out as soon as you can. I get extra credit points if @NeptuneHub decides to accept it by Friday! Thank you so much. I enjoyed working on this feature/bug so much.

NeptuneHub:
[...]

PR #436, opened 2026-04-18
fix: propagate MediaServerConnectionError instead of silently returning []

gulmaan87:
I was running an analysis and noticed something weird — when the media server 
credentials were wrong, the task still showed up as successful with the message 
" No new albums to analyze." That's confusing because it looks like everything 
worked fine when it actually couldn't even reach the server.

Dug into it and found the issue: all three providers (Jellyfin, Emby, Navidrome) 
were catching connection errors silently and returning an empty list. The analysis 
task had no way to tell the difference between "genuinely no albums" and "couldn't 
connect at all", so it just reported success either way.
What I changed:
Added a `MediaServerConnectionError` exception in `tasks/commons.py` so we have 
a proper signal for this case. Then updated Jellyfin, Emby, and Navidrome to raise 
it instead of swallowing the error. The analysis task now catches it and marks the 
job as FAILURE with a clear message telling the user to check their Setup page.

Also fixed the Setup page test-connection response — it was returning a hardcoded 
generic string instead of the actual error from the server, so users couldn't tell 
what went wrong. Now it passes the real error through. There was also a small typo 
("top player" → "top played") that I fixed while I was in there.
Pretty small change overall but it makes debugging a misconfigured server much less 
painful.

### DIFF
--- a/app_setup.py
+++ b/app_setup.py
@@ -262,7 +262,7 @@ def setup_api():
     except Exception as exc:
         app.logger.error('Setup save failed: %s', exc, exc_info=True)
         if is_test_connection:
-            return jsonify({'error': 'Unable to get top player song. Check the server log for details.'}), 500
+            return jsonify({'error': str(exc) or 'Unable to get top played song. Check the server log for details.'}), 500
         return jsonify({'error': 'Unable to save configuration. Check the server log for details.'}), 500
 
     response = make_response(jsonify({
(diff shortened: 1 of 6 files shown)


NeptuneHub:
Thanks but about error I would like to rework in the direction of a centralized error handling functionality that address this and similar issue in the code.

I'm going to close this to avoid duplicate effort. If you're interesting to contribute in a centralized error handling functionality please share your idea in the issue: https://github.com/NeptuneHub/AudioMuse-AI/issues/107

PR #439, opened 2026-04-18
feat: option to configure Flask listening address

L2jLiga:
Added environment variable to configure gunicorn bind address, so it can be changed to `[::]` for IPv6-only / dual-stack environments


NeptuneHub:
I don't have the interest right now to support IPV6, thanks.

PR #441, opened 2026-04-19
Migration - keep status on reload

NeptuneHub:
The migration provider functionality need improvement:

- If user run many migration, the `migration_session` table keep growing. This is wrong. At least one session should be there;
- When you reload the migration page, if a session is still running (in progress or dry_run status) it need to restore the session and give the user the possibility to proceed with the migration;
- When you reload the migration page, and only completed or in error session are there, a new session should be started and old one be deleted
- Credentail of an in progress session MUST NEVER come back in the frotend for security reason

### DIFF
--- a/app_provider_migration.py
+++ b/app_provider_migration.py
@@ -164,7 +164,11 @@ def provider_migration_page():
             row = cur.fetchone()
         if row:
             active_session_id = row[0]
-    except Exception:
+    except Exception as e:
+        logger.warning(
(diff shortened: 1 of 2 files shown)


NeptuneHub:
The resume mechanism should be generic enough, anyway I tested to refresh the page, in all different step, with this migration use case:

- Navidrome => Lyrion: OK
- Lyrion => Navidrome: OK
- Navirdome => Jellyfin: OK

No issue founded till now.

PR #444, opened 2026-04-19
Example on using an EnvironmentFile in the container files

p3ck:
Instead of repeating the same Environment entry in each container file. Unfortunately you need to specify the absolute path which may be different.


NeptuneHub:
Not interested in this, thanks.

Issue #215, opened 2025-12-04
[BUG] Analysis repeatedly attempts to process albums with "skipped" songs.


TJ09:
**Describe the bug**
Several albums in my library have "songs" that are either silent or too short to analyze. The analysis step correctly skips over these (e.g. `Skipping track [silence] by TV on the Radio as analysis returned None.`), but the next time analysis is run, these same albums are analyzed again, rather than being skipped as having complete analysis.

**To Reproduce**
0. Have an audio file that is completely silent. This can be easily generated and added to any album for testing.
1. Run analysis.
2. See that the album was launched and analyzed (log reports "analysis complete").
3. Run analysis again.
4. See that the album was again launched and analyzed.

**Expected behavior**
After the first analysis, subsequent runs are no-ops since all valid files have been analyzed (barring actual errors such as decoding failures).

**Environment (please complete the following information):**
 - Deployment: Docker Compose
 - AudioMuse-AI Version: v0.7.12-beta

**Additional Context**

To be clear, I don't expect any meaningful analysis to occur on short/silent songs, but re-analyzing the album on each run is wasted effort that ideally can be avoided.

NeptuneHub:
Thanks for the feedback. Sometimes happen that a songs is not analyze but a new analysis sovle the problem due to a momentaneus issue. If I mark a skipped songs as skip forever, it will never be analyzed again.

If are just few songs I think is not a big problem. If instead you found out multiple songs that are not possible to be analyzed I suggest to move in a different library and then use the `MUSIC_LIBRARIES` paramer as documented in the readme of this project to select only the library that you want to analyze. In this way you can say which library need to be analyzed and which not.

Maybe for the future a better implementation could be check in the code if the same song is skipped multiple time (let's say 3+?) to avoid both the cases. If you think it could be a good balance, I can keep in the enanchment for the future (touch the analysis is something that need extra attention).

TJ09:
> Maybe for the future a better implementation could be check in the code if the same song is skipped multiple time (let's say 3+?) to avoid both the cases. If you think it could be a good balance, I can keep in the enanchment for the future (touch the analysis is something that need extra attention).

Yeah, if something fails repeatedly I imagine that's unlikely to change (though perhaps it should be reset each version because it's possible a code change *does* fix it).

[...]

PR #450, opened 2026-04-21
Unanalyzable track UI feature 

camicasa:
Unanalyzable definition: During album analysis, a track that returns None upon analysis.

Changes to existing system:
	 - changed 'score' table to include analysis_status column
	 -  added database backfill method for existing users so their already 'analyzed' songs will be added to the analysis_status column
	 - added checks to make sure that an album with unanalyzable tracks will be skipped
	 - classified tracks as analyzed or unanalyzable 
	 - added UI where users can see what albums were skipped for unanalyzable reasons and which tracks were marked as unanalyzable when they press "Start analysis" on the main page
	 -  added logger info that goes into more detail, like when the track was marked as analyzable and what the title of it was, and what albums were skipped due to 'unanalyzable' tracks

Why? Link to bug: https://github.com/NeptuneHub/AudioMuse-AI/issues/215
Previously, albums containing tracks that could not be analyzed (e.g. silent or corrupt audio files) would be re-queued and re-processed on every analysis run. Since no meaningful data was written to the database for these tracks, the system had no memory of having attempted them, causing repeated wasted CPU time. With this fix, the system now remembers which tracks were unanalyzable and permanently skips the albums with them on future runs, eliminating the repeated processing and giving the user clear feedback on what was skipped and why.
Screenshot 2026-04-21 at 2 13 13 PM
Screenshot 2026-04-21 at 2 13 41 PM

How I tested it:
- Added a silent MP3 file as the only track in an album. Ran analysis once — confirmed the track was correctly marked as unanalyzable in the database, and the warning appeared in the logs and in the final Task Status message. Ran analysis again, confirmed the track and album were skipped immediately without re-attempting analysis, and the final Task Status message reflected the unanalyzable count.
- Added a silent MP3 file alongside regular songs in the same album. Ran analysis, confirmed the silent track was marked as unanalyzable while the regular tracks were successfully analyzed and marked as analyzed. Ran analysis again — confirmed that the album was skipped due to the 'unanalyzable' track and the UI correctly reflected that
- Confirmed that the analysis_status database migration correctly backfills existing analyzed tracks on first startup.
- Ran clustering and other task features after the fix — confirmed no regression in existing functionality.


tasks/analysis.py
```diff
@@ -1169,7 +1204,38 @@ def monitor_and_clear_jobs():
...
+
+                if unanalyzable_count == len(track_ids):
+                    logger.info(f"album name: '{album.get('Name')}', unanalyzable tracks: {', '.join(unanalyzable_names)}. ")
+                    albums_skipped += 1
+                    checked_album_ids.add(album['Id'])
+                    log_and_update_main(
+                        f"Skipping album '{album.get('Name')}' - all {unanalyzable_count} tracks previously marked as unanalyzable.",
+                        current_progress
+                    )
+                    continue
```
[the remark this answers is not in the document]
camicasa: Okay I can fix those three bugs right away

tasks/analysis.py
```diff
@@ -627,7 +627,15 @@ def get_existing_track_ids(track_ids):
...
-
+            
+            # Unanalyzable track feature - uncomment if you don't want to skip the album with unanalyzable tracks
+            # def get_unanalyzable_track_ids(track_ids):
+            #     if not track_ids: return set()
+            #     with get_db() as conn, conn.cursor() as cur:
+            #         track_ids_as_strings = [str(id) for id in track_ids]
+            #         cur.execute("SELECT item_id, title FROM score WHERE item_id IN %s AND analysis_status = 'unanalyzable'", (tuple(track_ids_as_strings),))
+            #         return {(row[0], row[1]) for row in cur.fetchall()}
+            #unanalyzable_tracks_set = get_unanalyzable_track_ids([str(t['Id']) for t in tracks])
```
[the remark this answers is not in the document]
camicasa: I agree.

tasks/analysis.py
```diff
@@ -656,6 +665,16 @@ def get_missing_mulan_track_ids(track_ids):
...
+                
+                # Unanalyzable track check
+                # Uncomment if you don't want albums skipped because of unanalyzable tracks
+                # track_id_str_ = str(item['Id'])
+                # track_title = item['Name']
+                # unanalyzable_ids = [track[0] for track in unanalyzable_tracks_set]
+                # if track_id_str_ in unanalyzable_ids:
+                #     unanalyzable_tracks_skipped_count += 1
+                #     logger.info(f"Skipping '{track_title}' - previously marked as unanalyzable.")
+                #     continue
```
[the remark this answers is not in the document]
camicasa: I agree.

camicasa:
I fixed all the issues that the Gemini bot raised. I also tested it

NeptuneHub:
Hi and thanks for the contribution. I’m very happy that AudioMuse-AI get your attention and you decided to contribute.

In case you would like also to chat with us you can reach on the Jellify discord server that host a AudioMuse-AI Chanel here:
- https://discord.gg/yf8fBatktn

Due to the importance of the analysis functionality and the possible impact I ask also to @rendyhd if you can give a look.

From my first code review I have some question/possible issue to raise:

- can you list all the test that you did? Did you tested also other connected functionality ? Because Analysis impact practically everything.
- when you analyze an album with ONLY some track to skip, are you sure you skip only those track and not the entire album?
- did you tested the new column creation on both legacy system and new deployment ?
- When you recreate all the index (Musicnn voyager index; Artist Index; 2D music map) what happen to the song that you add in the score table “as blank” and that also don’t have the corresponding embbeding table ?
- did you set this skipped only when the return is nine or also in exception case?
- What appen on the different search field that use score table? They will find a song even if skipped and then all the connected functionality will be in error because it doesn’t exist ?
- What if a track is not analyzed for temporary issue (network problem with the music server or similar) it will be skipped forever? Is it possible to set a grace period maybe of 3 analysis before skipping forever?
- The cleaning functionality how it works with this track that have an entry only in the score table ? I means when it’s a real orphan to be deleted ?

camicasa:
**can you list all the test that you did? Did you tested also other connected functionality ? Because Analysis impact practically everything.**
     - I tested all the other connected functionalities and everything was running smoothly. This feature shouldn't impact any other functionalities. It just gives a track an 'unanalyzable' tag. It's up to the user to delete that 'unanalyzable' track from its parent album so they can continue analyzing that album if they add more songs to the album. 
     
**when you analyze an album with ONLY some track to skip, are you sure you skip only those track and not the entire album?**
     - When there is an album with one newly 'unanalyzable' track it will analyze the track and mark the track as 'unanalyzable'. If that album is queued again for analysis that album will be skipped because there is AT LEAST one 'unanalyzable' track. This feature skips albums with 'unanalyzable' tracks, it doesn't skip those 'unanalyzable' tracks. If I have an album with an already marked as 'unanalyzable' track and I added a new track to the album that is a normal song, the analysis won't even attempt to analyze that new song because the album contains an 'unanalyzable' track and skip the album. The point of the feature is to relay the message to the user "hey this album has an unanalyzable track, you should take it out because everytime you start a "Start Analysis" task it will always queue the album for analysis because of that one unanalyzable track, and waste your time"  
     
**did you tested the new column creation on both legacy system and new deployment ?**
      - Sorry I did not. This is my first pull request so I didn't know how to test on legacy system? Like what does that even mean?
      -
**When you recreate all the index (Musicnn voyager index; Artist Index; 2D music map) what happen to the song that you add in the score table “as blank” and that also don’t have the corresponding embbeding table ?**
      - There is no way for an 'unanalyzable' song to not be added to the score table as blank. It gets added to the score table when it gets analyzed. It's the same process as before, the song reaches the point (If analysis is None; continue). The song gets added to the score table but there is no meaningful data to it. The only problem would be when an album with an existing 'unanalyzable' track is getting analyzed again with a new reg. song added to it. That new song wouldn't get added to the score table because the album got skipped before that point. That's why in the UI I have "Skipping album {album_name} has {unanalyzable_count} unanalyzable track(s): {unanalyzable_track_name}. Delete track(s) from album"
      
**did you set this skipped only when the return is nine or also in exception case?**
       - When a track that is getting analyzed returns None like before in the legacy code it would get skipped but with this feature it also gets marked as 'unanalyzable'
       
**What appen on the different search field that use score table? They will find a song even if skipped and then all the connected functionality will be in error because it doesn’t exist ?**       
       - There should be no error because even an 'unanalyzable' track will exist. It's analysis_status will just be 'unanalyzable'. Everything should be the same as before when you would skip a song. It just now has a tag to it. What happens now with a track that was skipped during analysis when it's looked up? because 99.99% sure it will be the same outcome for the 'unanalyzable' tracks

**What if a track is not analyzed for temporary issue (network problem with the music server or similar) it will be skipped forever? Is it possible to set a grace period maybe of 3 analysis before skipping forever?**
        - If a track makes it to the point "If analysis is None" and goes into the conditional then that is based on the analysis, ya know? Whatever is making that track not being analyzed like corrupted metadata is making the track analysis not work. Have users gone through that before? Where a network problem was outputting a perfectly analyzable song as "f"Skipping track {track_name_full} as analysis returned None." ? I think having a grace period would require more data collected on the track in the score table. We would have to keep track of the count and I don't think there is a reason for that. If a track is 'unanalyzable' there is something wrong with the track. The UI will let the user know and it's up to the user to figure out why their track is 'unanalyzable'. In the logs there is a reason why a song would be skipped. Maybe we could try and let the user know the specific reason why the track is bad according to the analyze_album function. 

**The cleaning functionality how it works with this track that have an entry only in the score table ? I means when it’s a real orphan to be deleted ?**
        - I'm sorry I'm confused by the question. Is there a cleaning functionality for these skipped tracks? Or should there be a cleaning functionality that deletes the 'unanalyzable' track?

app_helper.py
rendyhd: unclosed triple-quoted string

tasks/analysis.py
```diff
@@ -1179,6 +1195,35 @@ def monitor_and_clear_jobs():
...
+                    """, (tuple(track_ids),))
+                    unanalyzable_tracks = cur.fetchall()
+                if unanalyzable_tracks:
+                    unanalyzable_count = len(unanalyzable_tracks)
+                    unanalyzable_names = [t[1] for t in unanalyzable_tracks]
+                    albums_skipped += 1
+                    checked_album_ids.add(album['Id'])
+                    if unanalyzable_count == len(track_ids):
+                        logger.info(f"album name: '{album.get('Name')}', unanalyzable tracks: {', '.join(unanalyzable_names)}. ")
+                        albums_skipped += 1
```
rendyhd: duplicated

tasks/analysis.py
```diff
@@ -1179,6 +1195,35 @@ def monitor_and_clear_jobs():
...
+                        checked_album_ids.add(album['Id'])
```
rendyhd: duplicated

tasks/analysis.py
camicasa: oops yup. I just fixed that.

tasks/analysis.py
camicasa: just fixed

app_helper.py
camicasa: Yeah I just merged with that code that was incoming for your guys' main branch. I didn't check to see if there were errors. I fetched from upstream and it looks like there is a couple of bugs but I'll let the original author fix those. I don't want to break anything.

tasks/analysis.py
rendyhd: There were two right, albums_skipped += 1 and checked_album_ids.add(album['Id'])

app_helper.py
camicasa: I can also revert back to my original app_helper if you guys want to test it without the new code that was added the cleaning functionality i think

tasks/analysis.py
camicasa: Oh right. Sorry about that. I just pushed the new code without the duplicate checked_album_id and albums_skipped

app_helper.py
NeptuneHub: I triple-checked and main branch compile without issue, also if you look the app_helper.py fail in the main branch:

- https://github.com/NeptuneHub/AudioMuse-AI/blob/main/app_helper.py

The quote are open and closed correctly:
```

            # Create 'alchemy_anchors' table to persist named user anchors for reuse
            cur.execute("CREATE TABLE IF NOT EXISTS alchemy_anchors (id SERIAL PRIMARY KEY, name TEXT UNIQUE NOT NULL, centroid JSONB NOT NULL, created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP)")
            # Provider migration tool: wizard session state (one row per migration attempt)
            cur.execute("""
                CREATE TABLE IF NOT EXISTS migration_session (
                    id           SERIAL PRIMARY KEY,
                    created_at   TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
                    completed_at TIMESTAMP,
                    status       TEXT NOT NULL DEFAULT 'in_progress',
                    source_type  TEXT NOT NULL,
                    target_type  TEXT NOT NULL,
                    target_creds TEXT NOT NULL,
                    state        JSONB NOT NULL DEFAULT '{}'
                )
            """)
```

Could be that something went wrong with the merge? Please have a check.

camicasa: Okay i just pushed the correct version of app_helper with my analysis_status stuff. I also tested it and everything is still working smoothly. Your cleaning functionality should have no impact on the 'unanalyzable' feature

app_helper.py
```diff
@@ -890,7 +902,7 @@ def _parse_year_from_date(year_value):
         # Save analysis to score table
         cur.execute("""
             INSERT INTO score (item_id, title, author, tempo, key, scale, mood_vector, energy, other_features, album, album_artist, year, rating, file_path)
```
NeptuneHub: You miss to add analysis_status in the insert:

```INSERT INTO score (item_id, title, author, tempo, key, scale, mood_vector, energy, other_features, album, album_artist, year, rating, file_path, analysis_status)```

**Review by NeptuneHub — CHANGES_REQUESTED** [review]

File: app_helper.py (hunk 6 above)
Comment (camicasa) [reply app_helper.py]: Oops my bad. I must've missed that when I was added my code back into the new version of app_helper.py
I pushed a commit that fixes that. thanks for catching that

File: config.py
Comment (rendyhd) [inline config.py:1]: why did you change the clustering MIN?

Comment (camicasa) [reply config.py:1]: Because I didn't have the original minimum of 40 songs in my Navidrome library, and I wanted to test the clustering feature to make sure that it worked

**Comment by rendyhd** [comment]
Hi @camicasa, I've been thinking about the architectural approach and have some thoughts and suggestions that might drastically simplify te approach.

Before this PR, score had a single invariant: "a row in score means we successfully analyzed this track." The rest of the app relies on that invariant in half a dozen places. The PR silently breaks it by introducing a second kind of row that looks identical to a successful analysis except for one column's value. It's going to introduce so many checks.

I think it's cleaner to add a new table and not touch score: e.g.

```
  CREATE TABLE IF NOT EXISTS analysis_failures (
      item_id TEXT PRIMARY KEY,
      first_failure_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
      last_failure_at  TIMESTAMPTZ NOT NULL DEFAULT NOW(),
      failure_count    INT NOT NULL DEFAULT 1,
      last_reason      TEXT
  );
```

This would fix:
- score keeps its invariant. Zero changes needed to voyager rebuild, search, 2D map, artist index, orphan cleanup, dashboard, or the search_u trigger. The "which existing queries need a new filter?" question disappears because the answer is none.
- Grace period is a one-line predicate: WHERE failure_count < 3. TTL retry is OR last_failure_at < NOW() - INTERVAL '30 days'.
- Per-track skip is the natural default. In analyze_album_task iteration, check analysis_failures per track; that's where the per-track loop already is. Don't pre-check at the album level at all, or only skip the album when every track is in analysis_failures with exhausted budget.
- Failure metadata lives where it belongs: error messages, attempt history, operator notes. None of that belongs in a search/analysis table.
- Operator recovery is trivial: DELETE FROM analysis_failures WHERE item_id = ? (no risk of nuking a real row)
- No title corruption because there's no INSERT into score for failures.
- Migration is cheaper: CREATE TABLE IF NOT EXISTS only. No full-table UPDATE to rewrite every existing row in score.
- UI feedback is just as easy: JOIN analysis_failures af ON af.item_id = t.item_id when listing an album; or a count for the task summary.

What this means:
1. Revert config.py, deployment/docker-compose.yaml entirely from this PR - they don't belong to this feature.
2. Redesign around a dedicated analysis_failures table with failure_count and timestamps. Drop the analysis_status column from score and the skeleton-row INSERT.
3. Move the skip check into analyze_album_task's per-track loop so bad tracks are skipped individually. Album-level pre-skip is only defensible when every track's failure budget is exhausted, and even then it's an optimization, not a semantic requirement.
4. Add the retry budget @NeptuneHub asked for (e.g., MAX_ANALYSIS_RETRIES=3).

@NeptuneHub your thoughts?

NeptuneHub:
@camicasa please let's proceed with the @rendyhd review of an additional `analysis_failures `table.
It will solve some issue that I already see in the source code review (like search that return not analyzed song) as well as hidden issue in the idfferent rebuild logic.

Just one things that we need to take care in addition is the Migration functionality, that enable to migrate from one music server to another. In the best case scenario we will have a table doubled, in the worst case we will have collision.
When migration succeed I suggest to just clean the `analysis_failures` table.

In general better to try to re-analyze a corrupted song, you will lost some extra minutes in the nights, that having something that stop working.

Edit: Another nice to have, but I suggest of doing this in a next step in a new PR, will be a list of failed analysis song (maybe a new feature under the administration menu). So that user can know what is failing. It should be just a paginated view of this table.

camicasa:
Okay I’ll start working on it!

NeptuneHub:
@camicasa  Are you still working on this?

NeptuneHub:
Going to close due to no update on this.
Please raise a new PR when you're ready.

PR #451, opened 2026-04-22
Fix Navidrome migration dry-run ignoring session credentials

### LINKED ISSUE #449: [FEATURE] Provider migration - File Path mapping
When trying to migrate from one provider to another, if the same library is mapped on a new provider in a slightly different way (going from bare metal to docker, for example) it will fail to match even when expanding the search to match by title+artist

It would be great if we could say "hey in the existing file paths, replace part of it with this" ie: "/data/media/music/{{artist}}/{{album}} maps to /music/{{artist}}/{{album}}"

I could probably re-map my data in navidrome or just pull in the fresh analysis I did when I was planning this migration manually but this would provide more flexibility going forward

**Additional context**
Love that this feature was added!

rendyhd:
In migration mode, `mediaserver_navidrome.get_all_songs(user_creds=...)` called `_get_target_music_folder_ids()` unconditionally, which falls back to empty `config.NAVIDROME_*` globals when Navidrome is the target (not the live provider). That caused `getMusicFolders` to fail and the dry-run to return 0 songs.

When `user_creds` is supplied (migration-target probe), skip the `MUSIC_LIBRARIES` filter entirely. That filter reflects the source provider's folder selection and has no meaning for a migration target whose folder names may differ. Filtering post-migration stays available via the normal live-provider config path.

Regression tests added in `tests/unit/test_mediaserver.py` cover both branches: migration (filter skipped, creds forwarded to `search3`) and live-provider (filter still applied).

### DIFF
--- a/tasks/mediaserver_navidrome.py
+++ b/tasks/mediaserver_navidrome.py
@@ -256,8 +256,11 @@ def get_all_songs(user_creds=None):
     """
     Fetches all songs from Navidrome using admin or override credentials.
     If MUSIC_LIBRARIES is set, it will only return songs from those folders.
+    When user_creds is supplied (migration-target probe), the MUSIC_LIBRARIES
+    filter is skipped: it reflects the *source* provider's folder selection and
+    has no meaning on a migration target whose folder names may differ.
     """
-    target_folder_ids = _get_target_music_folder_ids()
+    target_folder_ids = None if user_creds else _get_target_music_folder_ids()
     
     # Case 1: Config is set, but no matching folders were found. Return no songs.
     if isinstance(target_folder_ids, set) and not target_folder_ids:
--- a/tests/unit/test_mediaserver.py
+++ b/tests/unit/test_mediaserver.py
@@ -1086,8 +1086,119 @@ class TestNavidromeGetRecentAlbums:
         mock_folders.return_value = set()  # Empty set = no matches
         
         albums = get_recent_albums(10)
-        
+
         assert albums == []
+        mock_request.assert_not_called()
+
+
+def _load_navidrome_module():
+    """Load tasks.mediaserver_navidrome directly, bypassing tasks/__init__.py.
+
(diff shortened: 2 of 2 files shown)


rendyhd:
Successfully tested with MUSIC_LIBRARIES set.
Fix is very simplified, during migration MUSIC_LIBRARIES is ignored for second provider, there's no point in building matching for it - any tracks that don't match don't get included anyway.

What we can do is a notification that user shouldn't forget to set their settings. Or making the UI match the Setup Wizard, I'm not sure yet. This fix will at least run if you migrate TO Navidrome with source MUSIC_LIBRARIES filters

NeptuneHub:
I did a source code review and seems that, at least for the matching of the song, Navidrome is the only impacted and the fix should solve the initial matching:

```
MUSIC_LIBRARIES filtering behavior
Navidrome
- uses _get_target_music_folder_ids() + getMusicFolders
- filter is applied in get_all_songs()
- PR patches this for migration probe

Jellyfin
- uses _get_target_library_ids() + /Library/VirtualFolders
- filter is applied in get_recent_albums() / album scans
- get_all_songs() does not use it, so migration is not affected

Emby
- uses _get_target_library_ids() + /emby/Library/VirtualFolders
- filter is applied in _get_recent_standalone_tracks() and get_recent_albums()
- get_all_songs() does not use it

Lyrion
- uses _get_target_paths_for_filtering() and path matching
- filter is applied in album/track scan code
- get_all_songs() bypasses it entirely, so migration is not affected
```

What I don't like is:
- We change a filter based on the presence of absence of credential: this for me is wrong. In future we can miss this details and introduce bug;
- In this way the migration finish, the user don't get allerted, and then the analysis start to fail because the music library don't match. 

For me a better fix should be:
IF MUSIC_LIBRARY is not empty, you SHOW an extra step (that othewise is hidden) where you let the user selecting the new music library. This will then update the final configuration at the end of the migration.

This will ends up probably in more change, because all music server will need to show the music library with credential, but  from an usability point of view seems better.

What do you think?

PR #455, opened 2026-04-23
Fix internal mapping of POSTGRES_PORT that incorrectly look at external port set in .env

knowgoodcoffee:
The problem: if the user changes the POSTGRES_PORT in the .env to a non-default value to avoid conflicts when other containers already are current using the default port, the AudioMuse-AI container will not start because the internal port mapping for the Flask and Worker services is also pointing to the external value defined in the .env, even though postgres is listening internally to the default 5432. 

For example, the user changes the POSTGRES_PORT in .env to something like "5431", the Flask and Worker services will start looking _internally_ to 5431 while postgres is still internally on the same 5432 port, as it's not being mapped internally to any other port (as seen in line 22). This will cause connection failed errors in the logs and AudioMuse-AI will fail to start. Thus, these services only need to point to the internal 5432 as Docker is managing the mapping. 

If the port is change in the .env and there happens not to be any conflicting services, it's likely AudioMuse-AI will start seemingly normally, but fail to connect behind the scenes, causing confusing errors.


NeptuneHub:
Please address the comment from Gemini if you are still interested in get the PR merged and then mark as ready to be reviewed. Thanks.

@rendyhd because you're the mast&command of docker test, when this PR will be ready to be review can you have a check? then if it's ok for you I'll merge.

rendyhd:
Sure. @knowgoodcoffee let me know when you're good to go

knowgoodcoffee:
Thank you @NeptuneHub and @rendyhd. I just made the changes.

NeptuneHub:
The change you did was almost correct:
- is correct that you DON'T change the postgres/redis internal PORT, but only the external one. There is no need to change the internal one due to the fact that docker automatically create internal network.
- wasn't correct delete the `REDIS_URL` parameter, I reintroduced because could be useful to re-use an external redis: most of the user will not need it, but for example I use it for testing purposes

Thanks for your contribution and for spotting this error.

NeptuneHub:
I partially reverted and reworked this PR. My main point is that internal configuration values should remain unchanged. In this example, where Redis and PostgreSQL are also deployed, their URLs should not be modified either.

For a basic deployment, the user should only need to adjust the host port if it is already in use, as well as the database username and password for security reasons. Adjusting the time zone (TZ) is optional.

For advanced users who want to reuse an existing database or Redis instance, this example is not intended to cover those scenarios. As advanced users, they are expected to determine the best approach based on their own architecture.

Thanks for your effort and for identifying the issue.

knowgoodcoffee:
Thank you! The only issue I really was trying to address in this PR was just the external port configuration of Postgres and the inadvertent internal network change, but the Gemini code bot was so insistent (as well as the LLM in my own testing) to change Redis as well... so I eventually gave in.

PR #481, opened 2026-05-04
Refactor docker-compose with YAML anchors

freekngdom:
Added YAML anchors for reusable configuration blocks and improved service definitions.


NeptuneHub:
Hi, I appreciate your interest in the project, but I prefear to keep this example simple.

It's not my purpose to have extensive effort on the specific way to deploy AudioMuse-AI that instead should be up to the final user.

Thanks.

PR #483, opened 2026-05-04
FIx jellyfin lyrics api

NeptuneHub:


corentinaltepe:
Thank you for this fix. I was about to report the bug since I just noticed songs with lyrics available were processed with Whisper anyway. I disabled lyrics analysis until this fix is released to avoid using too much energy.

Issue #150, opened 2025-10-29
[BUG] Jellyfin Fingerprint playlist generation naming scheme not consistent


Th3K1ngP1n:
**Describe the bug**
When creating fingerprint playlist is gernerated by jellyfin, the name of the playlist is followed by a numer 

for example:

Name-Fingerprint ● 83

next generated one will be 

Name-Fingerprint ● 84

and so on.

This cause issues with Symfonium online playlist, as the playlist is assumed different each time it is generated. 
I cannot say that the naming scheme is the only thing preventing symfonium from having online playlist, maybe the ID also. But other online playlists work. 

**To Reproduce**
0. Load fingerprint playlist online first to symfonium
1. Generate a new fingerprint playlist
2. Sync symfonium
3. Original Playlist is not empty and gone to limbo

**Expected behavior**
Old playlist is replaced with contents of new playlist without changing the playlist id or naming scheme. 

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. Debian, Windows]
 - Deployment: [e.g. Docker Compose, K3s, Helm]
 - AudioMuse-AI Version: [e.g. v0.4.0-beta, latest]
 - Jellyfin/Navidrome Version: [e.g. Navidrome 0.57.0]
 - Browser: [e.g. chrome, safari]
 - CPU: [e.g. Intel I5-6500, Arm xyz]
 - RAM: [e.g. 8GB DDR4]
 - Disk: [e.g. SSD Nvme]

**Additional context**
Add any other context about the problem here. For example, logs from the `audiomuse-ai-worker` or `audiomuse-ai-flask` containers can be very helpful. Also logs from the Browser could help too (on Chrome > right mouse button > click on inspect > go to Network tab > click on call that is in red/fail and copy the content).
If the bug is related to a specific album/songs please share information like format, size, bitrate. On linux you can use `mediainfo yoursong.flac`

NeptuneHub:
I don't understand the bug.
When you are in AudioMuse-AI front-end you can put the name that you want and it should be visible everywhere like any other playlist.
When you use AudioMuse-AI jellyfin plugin, that create it automatically, it should delete the old one playlist and create a new one. This because it assign the exact same name and delete the other.

Worst case scenario, if plugin is not able to delete the old one, you should have two fingerprint playlist with the same name (I look that this same tames could happen). This because Jellyfin accept two or more playlist with the same name without issue.

Can you explain bettwe in which of this scenario you are?
[...]

PR #501, opened 2026-05-07
Feature #336: cron Sonic Fingerprint reuses one stable playlist

### LINKED ISSUE #336: [FEATURE] Sonic Mix single Playlist for Navidrome
Hello, I would like to see the same feature as https://github.com/NeptuneHub/AudioMuse-AI/issues/150 for Navidrome. In short, recreate playlist content but not playlist item. This allows auto syncing to Symfonium with online playlist first. 

Same as the one for Jellyfin.

**Additional context**
I think its fairly self explanatory. If you need any more details I am happy to provide.

rendyhd:
## Problem

  The scheduled `sonic_fingerprint` cron task (`app_cron.py`) called
  `create_playlist_from_ids("Sonic Fingerprint (Cron YYYY-MM-DD)", ids)`,
  which routes through `create_instant_playlist` (suffixed `_instant`) and
  creates a brand-new server-side playlist on every run.

  For "online-first" sync clients on Navidrome (Symfonium etc.) this breaks
  the user's saved sync because the server-side playlist ID changes each
  cron tick. Reported in #336; the parallel issue on Jellyfin (#150) was
  fixed inside the Jellyfin plugin, not in core, so all four core backends
  were still affected.

  ## Approach

  Per @NeptuneHub's direction in the issue thread:
  > "ad hoc playlist creation functionality only for fingerprint that then
  > need also to be reimplemented on all the music server. The logic here
  > should be to create the playlist with a specific name only if a previous
  > one doesn't exist, otherwise update the previous one."

  Add one new dispatcher function and one new helper per provider; existing
  `create_instant_playlist` / `create_playlist` / `create_playlist_from_ids`
  flows are not touched, so UI-driven and chat-driven playlist paths are
  unaffected.

  app_cron.run_due_cron_jobs (sonic_fingerprint branch only)
    └─ generate_sonic_fingerprint()
    └─ if results empty → log + skip (preserve previous playlist)
    └─ mediaserver.create_or_replace_playlist(name, track_ids)  [NEW]
     ──└─ provider_create_or_replace_playlist(name, ids)      [NEW per provider]
     ────├─ get_playlist_by_name(name)
     ────├─ exists  → clear-then-add (batched), preserve same playlist ID
     ────└─ missing → create-then-add-rest (batched)

  A new `SONIC_FINGERPRINT_CRON_PLAYLIST_NAME` config constant
  (env-overridable, default `"Sonic Fingerprint by AudioMuse-AI"`) replaces
  the date-suffixed name. MPD is intentionally out of scope; the dispatcher
  raises `NotImplementedError` for MPD and the cron handler catches it and
  falls back to the legacy date-suffixed `create_playlist_from_ids` path so
  MPD users see no regression.

  ## Per-backend implementation

  - **Navidrome**: `getPlaylist` + batched `updatePlaylist` with descending
    `songIndexToRemove` (40/batch) to clear, then `_add_to_playlist` for new
    ids. Preserves the playlist ID across runs.
  - **Jellyfin**: GET/DELETE/POST on `/Playlists/{Id}/Items` in batches of
    100, using each item's `PlaylistItemId` for `entryIds`. Preserves the
    playlist ID.
  - **Emby**: same shape under `/emby/` with capitalized params. Preserves
    the playlist ID.
  - **Lyrion**: `delete_playlist` + `_create_playlist_batched` when an
    existing playlist is found. Lyrion's underlying `_add_to_playlist`
    primitive is destructive (load → add → delete original → save), so
    the numeric playlist ID may change across runs — documented limitation.
    Lyrion clients track playlists by name, which remains stable.

  ## Commits

  1. `Fix #336: stable single playlist for cron sonic_fingerprint across providers`
     — main feature: dispatcher, four backend implementations, cron handler change, tests.
  2. `Fix #336: don't silently report success when upsert partially fails`
     — review fixes: return None when add fails after a successful clear (Navidrome /
     Jellyfin / Emby), check `delete_playlist` return value on Lyrion to avoid
     creating duplicates, drop dead Id/id fallbacks, consolidate dispatcher raises.
  3. `Fix: jellyfin get_playlist_by_name was returning the wrong playlist`
     — Jellyfin's `/Users/{userId}/Items?Name=...` silently ignores the Name filter
     and returns every playlist. The old `playlists[0]` returned a wrong/stale
     playlist (mirrors a workaround the Emby version has had since day one).
  4. `Fix: lyrion delete_playlist mis-treated empty-dict success as failure`
     — LMS returns `{}` on successful `playlists delete`; the existing `if response`
     check evaluated that as falsy and reported failure, causing the upsert to
     abort and leaving Lyrion users with disappearing playlists.

  ## Tests

  - `tests/unit/test_mediaserver.py` — new test classes for each provider
    covering missing-playlist (create), existing-playlist (replace, ID
    preserved), batching across the per-provider batch size, empty
    `item_ids` early-return, and add-failure-returns-None semantics.
  - `tests/unit/test_app_cron.py` (new) — cron sonic_fingerprint branch:
    empty-results skip, non-empty calls upsert with the constant name, MPD
    fallback to legacy `create_playlist_from_ids`.

  `python -m pytest tests/unit/test_mediaserver.py tests/unit/test_app_cron.py` →
  **125 passed**.

  ## Test plan

  - [x] Unit tests pass (125/125 in the affected files)
  - [x] Live smoke on Navidrome — first run creates, subsequent runs replace
        contents with same playlist ID; Symfonium-style ID-tracking clients
        remain stable (confirmed playlist changes)
  - [x] Live smoke on Jellyfin — same Id preserved across cron ticks (confirmed playlist changes)
  - [x] Live smoke on Emby — same Id preserved across cron ticks (confirmed playlist changes)
  - [x] Live smoke on Lyrion — upsert completes cleanly; numeric ID flaps as
        documented (name remains stable). Creates new playlist with same Name. (didn't test if it changes content because my Lyrion instant didn't report plays)
  - [x] Empty-fingerprint path preserves the previous playlist

### DIFF
--- a/tasks/mediaserver_jellyfin.py
+++ b/tasks/mediaserver_jellyfin.py
@@ -10,7 +10,7 @@ from tasks.mediaserver_helper import detect_path_format
 logger = logging.getLogger(__name__)
 
 REQUESTS_TIMEOUT = 300
-JELLYFIN_PLAYLIST_BATCH_SIZE = 50
+JELLYFIN_PLAYLIST_BATCH_SIZE = 100
 
 # ##############################################################################
 # JELLYFIN IMPLEMENTATION
@@ -591,18 +591,13 @@ def _get_playlist_entry_ids(playlist_id):
 
 
 def _remove_playlist_entries(playlist_id, entry_ids):
-    """DELETEs entries from a Jellyfin playlist in batches. Returns True on full success.
-
-    ``userId`` must be passed as a query param: Jellyfin's DELETE handler can't resolve
-    a user from an API token alone and otherwise 400s with "Guid can't be empty".
-    See jellyfin/jellyfin#12999, #13476, #14910.
-    """
+    """DELETEs entries from a Jellyfin playlist in batches. Returns True on full success."""
     if not entry_ids:
         return True
     url = f"{config.JELLYFIN_URL}/Playlists/{playlist_id}/Items"
     for i in range(0, len(entry_ids), JELLYFIN_PLAYLIST_BATCH_SIZE):
         batch = entry_ids[i:i + JELLYFIN_PLAYLIST_BATCH_SIZE]
-        params = {"entryIds": ",".join(batch), "userId": config.JELLYFIN_USER_ID}
+        params = {"entryIds": ",".join(batch)}
         try:
             r = requests.delete(url, headers=config.HEADERS, params=params, timeout=REQUESTS_TIMEOUT)
             r.raise_for_status()
--- a/tasks/mediaserver_jellyfin.py
+++ b/tasks/mediaserver_jellyfin.py
@@ -657,7 +657,7 @@ def create_or_replace_playlist(playlist_name, item_ids, user_creds=None):
         logger.info(f"✅ Jellyfin: created playlist '{playlist_name}' (Id={new_id}) with {len(item_ids)} tracks")
         return {**created, 'Id': new_id, 'Name': created.get('Name', playlist_name)}
 
-    playlist_id = existing.get("Id") or existing.get("id")
+    playlist_id = existing.get("Id")
     if not playlist_id:
         logger.error(f"Jellyfin create_or_replace_playlist: existing playlist '{playlist_name}' has no Id")
         return None
@@ -670,7 +670,9 @@ def create_or_replace_playlist(playlist_name, item_ids, user_creds=None):
         return None
 
     if not _add_items_to_playlist(playlist_id, item_ids):
+        # Items were already cleared above; signal failure so the cron handler doesn't log success.
         logger.error(f"Jellyfin create_or_replace_playlist: failed to add tracks to playlist {playlist_id}")
+        return None
 
     logger.info(f"✅ Jellyfin: replaced contents of playlist '{playlist_name}' (Id={playlist_id}, tracks={len(item_ids)})")
     return {**existing, 'Id': playlist_id, 'Name': existing.get('Name', playlist_name)}
--- a/tasks/mediaserver_jellyfin.py
+++ b/tasks/mediaserver_jellyfin.py
@@ -10,7 +10,7 @@ from tasks.mediaserver_helper import detect_path_format
 logger = logging.getLogger(__name__)
 
 REQUESTS_TIMEOUT = 300
-JELLYFIN_PLAYLIST_BATCH_SIZE = 100
+JELLYFIN_PLAYLIST_BATCH_SIZE = 50
 
 # ##############################################################################
 # JELLYFIN IMPLEMENTATION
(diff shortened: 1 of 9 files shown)


tasks/mediaserver_jellyfin.py
```diff
@@ -555,3 +564,124 @@ def create_instant_playlist(playlist_name, item_ids, user_creds=None):
...
+            logger.error(f"Jellyfin create_or_replace_playlist: create failed for '{playlist_name}': {e}", exc_info=True)
+            return None
+
+        new_id = created.get("Id")
+        if not new_id:
+            logger.error(f"Jellyfin create_or_replace_playlist: created '{playlist_name}' but response had no Id")
+            return None
+
+        if rest and not _add_items_to_playlist(new_id, rest):
+            logger.error(f"Jellyfin create_or_replace_playlist: created '{playlist_name}' but failed to add overflow tracks")
```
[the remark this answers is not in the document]
rendyhd: Navidrome behaves the same way, current code is consistent, asymmetry within each backend (replace=strict, create=lenient) is intentional because the create branch leaves the user with partial-but-useful state, while the replace branch leaves an empty playlist if it fails

rendyhd:
@NeptuneHub this one was done and fully tested

tasks/mediaserver_jellyfin.py
NeptuneHub, line 1: I just set a cron job thar run each 1 minutes on Jellyfin.

The first run created the playlist
The second run return this error in the flask container:

```

│ [INFO]-[08-05-2026 11-50-46]-Total unique songs in final fingerprint playlist: 100                                                                                                                            │
│ [ERROR]-[08-05-2026 11-50-47]-Jellyfin _remove_playlist_entries: batch starting at 0 failed for playlist df295f22a1dd1ce80d6f131457c1c4e4: 400 Client Error: Bad Request for url: http://jellyfin.192.168.3.1 │
│ Traceback (most recent call last):                                                                                                                                                                            │
│   File "/app/tasks/mediaserver_jellyfin.py", line 603, in _remove_playlist_entries                                                                                                                            │
│     r.raise_for_status()                                                                                                                                                                                      │
│   File "/usr/local/lib/python3.12/dist-packages/requests/models.py", line 1028, in raise_for_status                                                                                                           │
│     raise HTTPError(http_error_msg, response=self)                                                                                                                                                            │
│ requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://jellyfin.192.168.3.169.nip.io:8087/Playlists/df295f22a1dd1ce80d6f131457c1c4e4/Items?entryIds=0dff55c130d348fba1e34bd505505e9c%2C │
│ [INFO]-[08-05-2026 11-50-47]-Cron: upserted 'Sonic Fingerprint by AudioMuse-AI' (playlist_id=None, tracks=100, job_id=5b738884-4bc0-4c86-a371-c9e1379646e5)                                                   │
│ [INFO]-[08-05-2026 11-50-47]-Cron: ran sonic fingerprint synchronously (job_id=5b738884-4bc0-4c86-a371-c9e1379646e5)          
```



Please check and fix on jellyfin and do similar test on all the other music server. Thanks.

NeptuneHub (changes requested):
Need to check the issue on jellyfin first.

tasks/mediaserver_jellyfin.py
```diff
@@ -675,7 +686,18 @@ def create_or_replace_playlist(playlist_name, item_ids, user_creds=None):
         return None
 
     if not _remove_playlist_entries(playlist_id, entry_ids):
```
NeptuneHub: This is not the proper way to do it. We already know that on some version of Jellyfin it is not supported. You need to use a TRY & CATCH side in order to handle the error with a warning.

Instead in this way you have both:
- the full hunandled error with the traceback, that could also allarm the user
```
 [ERROR]-[08-05-2026 15-56-26]-Jellyfin _remove_playlist_entries: batch starting at 0 failed for playlist df295f22a1dd1ce80d6f131457c1c4e4: 400 Client Error: Bad Request for url: http://jellyfin.192.168.3.1 │
│ Traceback (most recent call last):                                                                                                                                                                            │
│   File "/app/tasks/mediaserver_jellyfin.py", line 603, in _remove_playlist_entries                                                                                                                            │
│     r.raise_for_status()                                                                                                                                                                                      │
│   File "/usr/local/lib/python3.12/dist-packages/requests/models.py", line 1028, in raise_for_status                                                                                                           │
│     raise HTTPError(http_error_msg, response=self)                                                                                                                                                            │
│ requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://jellyfin.192.168.3.169.nip.io:8087/Playlists/df295f22a1dd1ce80d6f131457c1c4e4/Items?entryIds=0dff55c130d348fba1e34bd505505e9c%2C │
```
- and just after your warning:

```
│ [WARNING]-[08-05-2026 16-00-32]-Jellyfin: reuse of existing playlist 'Sonic Fingerprint by AudioMuse-AI' (Id=df295f22a1dd1ce80d6f131457c1c4e4) didn't work — in-place item removal failed (likely Jellyfin <  │
```

Please handled this error correctly.

NeptuneHub (changes requested):
Need better errro handling with Try&catch

PR #505, opened 2026-05-07
Feat/setup wizard redesign

rendyhd:
Been playing around with a setup redesign, mainly grouping and adding dropdowns / toggles.
Screenshot 2026-05-07 215405
Screenshot 2026-05-07 215416
Screenshot 2026-05-07 215422
Screenshot 2026-05-07 215428
Screenshot 2026-05-07 215433
Screenshot 2026-05-07 215438
Screenshot 2026-05-07 215443
Screenshot 2026-05-07 215452

### DIFF
--- a/static/setup-wizard.css
+++ b/static/setup-wizard.css
@@ -1,0 +1,534 @@
+/* ============================================================================
+   AudioMuse-AI — Setup Wizard
+   Sticky-rail layout, grouped cards, search, sticky save bar.
+   Inherits design tokens from style.css (--bg-card, --text-main, etc.).
+   ============================================================================ */
+
+/* Wizard sits inside .container; .container is already centered + max-width: 960px.
+   We widen on the setup page only so the rail + content fit comfortably. */
+.setup-wizard {
+    --wizard-rail-width: 18rem;
+    --wizard-gap: 1.25rem;
+    --wizard-card-radius: 1rem;
+    --wizard-savebar-height: 4.5rem;
+    margin: 0 auto;
+    max-width: 1200px;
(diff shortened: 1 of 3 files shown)


NeptuneHub:
I like the idea to organize that section, the advanced parameter section was a workaround, but I have some point:

1. the additional menu is not suitable for smartphone use, also I don't like the idea that you have double menu and this doesnt match the actual graphics. Will be better the list of voice collapsed where the user click to expand, being sure that everything is correctly visible on a smartphone.
2. we need to be sure that a new parameter automatically appear in advanced parameter if NOT excluded if NOT specified elswhere.
3. in a redesign of the setup another importat things that I would like to se is true/false value predefined so user don't have to type them
4. is important that the logic of "send only changed parameter" is preserved

Also we need to be sure that the source code remain simple and manageble.

rendyhd:
Not sure if I follow all the comments. This was just a brainstorm, I'll leave it to what you have in mind.

Regarding mobile, idea was that on mobile the menu would move to the top.  
image

NeptuneHub:
The horizontal menu is not the best, as you see you have the butotn of the other menu that go over, and the link of the other menu that go out. Also having 2 menu at all is not th ebest.

I'm thinking just a page with something like:
First line title: Media Server
Second line description: Connection parameter

it start collapsed but if you click it expand with all the voice. This for all the voice, and when you expan one, the other get collapsed.

This is good to view on both a desktop and a mobile screen and avoid to have multiple menu in one screen

rendyhd:
Closing this because we'll go in a different direction

PR #580, opened 2026-05-27
Add /api/sync endpoint for mobile library sync

### LINKED ISSUE #507: [FEATURE] /api/sync endpoint for mobile sync
Hi @NeptuneHub, as you might remember I've been working on a mobile player since the start of the year. I have this idea of a player that is different from others - it doesn't work from a library perspective but is build around the data we capture in AudioMuse-AI. For offline playback and local performance I'd really need a sync endpoint that I would like to build.

## Summary

Add a new GET /api/sync REST endpoint that exports the AudioMuse-AI analysis library as paginated JSON, so the app can mirror it into a local database for offline browsing and on-device algorithms.

## Requirements

Endpoint
- GET /api/sync — auth required by default (Bearer API_TOKEN or JWT cookie); follows AudioMuse-AI's existing /api/* auth pattern. Honors AUTH_ENABLED=false for opt-out.
- Query params: since (ISO 8601, incremental), include_embeddings (bool, default true), page (default 1), limit (default 500, max 1000)
- Status codes: 200 success, 400 bad since, 401 {"error":"Unauthorized"}, 500 server error

Response format
```
{
  "tracks": [
    {
      "id": "abc123-def456",
      "title": "Echoes",
      "artist": "Pink Floyd",
      "album_artist": "Pink Floyd",
      "album": "Meddle",
      "year": 1971,
      "tempo": 117.5,
      "key": "C# Minor",
      "scale": "minor",
      "mood_vector": "progressive rock:0.92,psychedelic rock:0.85,art rock:0.78,ambient:0.72,experimental:0.65",
      "energy": 0.07,
      "other_features": "relaxed:0.8,atmospheric:0.6",
      "rating": 5,
      "umap_x": 12.34,
      "umap_y": -5.67,
      "updated_at": "2026-03-15T14:30:00+00:00",
      "embedding": "<base64-encoded float32[200]>",
      "clap_embedding": "<base64-encoded float32[512]>"
    }
  ],
  "deleted_ids": ["xyz789-deleted-id"],
  "total_tracks": 15000,
  "model_version": "musicnn_v1.1.2",
  "provider_type": "navidrome",
  "has_more": true,
  "next_page": 2
}
```

## Field notes
- id — string, equal to score.item_id (mediaserver GUID). Mobile uses it as local PK and to build stream URLs (Navidrome /rest/stream, Jellyfin /Items/{id}/Audio).
- artist — comes from score.author (renamed in response)
- energy — raw 0.01–0.15, not normalized (mobile normalizes client-side)
- embedding — 200-D MusiCNN, base64 float32 little-endian
- clap_embedding — 512-D CLAP, base64 float32; omitted when CLAP_ENABLED=false
- embedding and clap_embedding keys are fully omitted (not null) when include_embeddings=false
- provider_type on envelope — single-mediaserver per install (jellyfin | navidrome | lyrion | emby); applies to every track. Mobile wipes local DB and full-resyncs
on mismatch.
- model_version — mobile triggers full re-sync when this changes between syncs
- deleted_ids — string[] of item_ids; empty [] when since is omitted; same array repeated on every page

## Schema additions
- New score.updated_at column (auto-bumped by trigger)
- New deleted_tracks table (item_id TEXT, deleted_at TIMESTAMP)
- Triggers: score_updated_at_trigger and score_deletion_logger

## Behavior
- Provider migration fires triggers normally — no special bypass. Mobile detects provider_type change and full-resyncs anyway, so trigger fallout is irrelevant.
- Bad since → 400 {"error": "Invalid 'since' parameter. Use ISO 8601 format."}

rendyhd:
Paginated read-only export of the analysis library: track metadata, mood/energy, MusiCNN + CLAP embeddings (base64), UMAP coords, and deletion tombstones for incremental sync.

- New score.updated_at column + BEFORE INSERT/UPDATE trigger for ?since= incremental deltas
- New deleted_tracks table + AFTER DELETE trigger for deletion log
- /api/sync route: mpd -> 501, page/limit clamping (cap 1000), conditional embedding/CLAP joins, naive-UTC since normalization
- 32 unit tests (mpd gate, since parsing, envelope shape, pagination math, CLAP toggle, base64 roundtrip, UMAP lookup, DB error -> 500)

## AS-IS
does not exist

## TO-BE
Adds a read-only `GET /api/sync` endpoint, a paginated bulk export of the analysis library for companion/client apps:

- **Track payload:** metadata (title, artist, album_artist, album, year, tempo, key, scale, rating), `mood_vector`, `energy`, `other_features`, UMAP 2D coords (`umap_x`/`umap_y`), and base64 MusiCNN + CLAP embeddings.
- **Incremental sync:** `?since=<ISO 8601>` returns only tracks changed since that time (new `score.updated_at` column maintained by a BEFORE INSERT/UPDATE trigger), plus `deleted_ids` — deletion tombstones from a new `deleted_tracks` table populated by an AFTER DELETE trigger on `score`.
- **Pagination:** `page` / `limit` (capped at 1000); envelope reports `total_tracks`, `has_more`, `next_page`.
- **Toggles:** `include_embeddings=false` skips the embedding joins/payload; CLAP fields are included only when CLAP is enabled.
- **Guard:** `mpd` media server returns `501` (not yet supported).

Response envelope:
```json
{ "tracks": [...], "deleted_ids": [...], "total_tracks": N, "provider_type": "...", "has_more": false, "next_page": null }
```
All changes are additive and idempotent, the new column/table/triggers are created in init_db() with IF NOT EXISTS, so existing installs migrate transparently on startup.

## Test
Unit (tests/unit/test_api_sync.py, 32 tests — all pass): 
pytest tests/unit/test_api_sync.py
Covers: mpd→501 gate, since parsing (Z and +offset normalized to naive-UTC), envelope shape, pagination math, include_embeddings toggle, base64 embedding
round-trip, UMAP lookup (hit/miss), deleted_ids gating (only on ?since=), and DB error → 500.

Manual (NVIDIA image, Navidrome backend, ~16k analyzed tracks):
# full sync
curl 'http://127.0.0.1:8000/api/sync?include_embeddings=true&page=1&limit=10'           # → 200
# incremental sync
curl 'http://127.0.0.1:8000/api/sync?since=2026-01-01T00:00:00Z&include_embeddings=true&page=1&limit=10'  # → 200
Verified the envelope, total_tracks, pagination, base64 embeddings, and a well-formed deleted_ids array.

## Other useful information
- Read-only endpoint (no writes on the request path); provider-agnostic (reads from the score analysis table), tested with Jellyfin.
- Embeddings are base64-encoded little-endian float32 arrays — MusiCNN 200-d, CLAP 512-d.
- deleted_ids are mediaserver GUIDs (score.item_id) — the same value returned as each track's id — so a client can delete the matching local row.
- The since filter uses score.updated_at; on first deploy the column defaults to CURRENT_TIMESTAMP, so the initial incremental window starts at deploy time.

## Checklist

**Type of change:**
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [ ] Intel
- [ ] ARM
- [ ] -noavx2
- [x] NVIDIA image

**Tested on media server:**
- [ ] Jellyfin
- [x] Navidrome
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [x] API Documentation
- [x] Unit test
- [ ] Integration test

**Other:**

**Related ISSUE:** Closes #507


rendyhd:
@NeptuneHub this one is ready to be reviewed, been running it for a while now.

NeptuneHub:
Hi,
I'll be happy to add the new /api/sync endpoint, in read only, in order to support yours and other external app that maybe want to run some offline mode (I suppose that this is the use case). The only change here that I want to ask is to max the paginate to 500 instead of 1000.

**My big concern is instead on the database side.**  The PR touch the score table, that is a central one in AudioMuse-AI and not only to add a column but also to add some trigger that run on every analysis, cleaning and migration. Not talking about that deleted_tracks just keeps growing.

**I prefear if you remove all the change to the table (both new column and triggers), since are not used to audiomuse-ai itself.**

AudioMuse-AI can export the data but the logic than need to run on the phone part.

Why you don't think it different ? You really need ALL the audiomuse-ai information on a phone? Maybe export only the index and his metadta is already enough to run the offline functionality on the phone. Instead for online functionality you don't need to export data and you can call the API itself.

app_helper.py
NeptuneHub, line 1: I want to avoid extra column and trigger on AudioMuse-AI database for functionality not directly used from AudioMsue-AI itself.

app_sync.py
NeptuneHub, line 1: Please reduce the max pagination from 1000 to 500.

NeptuneHub (changes requested):

app_helper.py
rendyhd: Removed all changes to the database and triggers.

app_sync.py
rendyhd: For payload reduced to 500, for index only left it at 1000

rendyhd:
> Hi, I'll be happy to add the new /api/sync endpoint, in read only, in order to support yours and other external app that maybe want to run some offline mode (I suppose that this is the use case). The only change here that I want to ask is to max the paginate to 500 instead of 1000.
> 
> **My big concern is instead on the database side.** The PR touch the score table, that is a central one in AudioMuse-AI and not only to add a column but also to add some trigger that run on every analysis, cleaning and migration. Not talking about that deleted_tracks just keeps growing.
> 
> **I prefear if you remove all the change to the table (both new column and triggers), since are not used to audiomuse-ai itself.**
> 
> AudioMuse-AI can export the data but the logic than need to run on the phone part.
> 
> Why you don't think it different ? You really need ALL the audiomuse-ai information on a phone? Maybe export only the index and his metadta is already enough to run the offline functionality on the phone. Instead for online functionality you don't need to export data and you can call the API itself.

I've removed all the db changes.

Why I sync all the track data:
At first it was for offline playback, both fully offline (airplane mode) and to account for folks not exposing their AudioMuse-AI server to the outside world. But during tested I've also been experimenting with a lot of different ways on how we could use it to do discover, trying different recommendation algorithms and tweaking them, many failures but some fun ones.

NeptuneHub (approved):

PR #595, opened 2026-05-31
Feat/Add MAX_SONGS_PER_ALBUM diversity cap (mirrors MAX_SONGS_PER_ARTIST)

### LINKED ISSUE #525: [FEATURE] Sonic Fingerprint cron job: limit number of songs from same Album
I listen to a lot of DJ mixes, where an album consists of several tracks mixed by a DJ. Naturally, these songs will have a sonic similarity, due to the nature of Djing where you want a degree of similarity between tracks. When I create a sonicfingerprint mix, I often has over 10 songs from the same Dj album due to the sonic similarity. Also non DJ mix albums sometimes have a high similarity between tracks.

Limit the amount of tracks from same album. For example no more than 2. 

Ive checked the parameters but didnt find anything to limit this.

rendyhd:
## AS-IS
No max album for sonic fingerprint

## TO-BE
Adds a per-album cap analogous to MAX_SONGS_PER_ARTIST, enforced across the same sites: similarity (Voyager vector/id search + greedy walk), clustering, song paths, CLAP text search, lyrics search, and SemGrove. Off by default (0 = disabled) so existing playlists are unchanged.

Also caps same-album tracks in Sonic Fingerprint, including the top-played seed songs (which bypass the in-search caps). The cap is enforced globally across seeds + neighbors so the playlist truly respects the limit; neighbors are over-requested when the cap is active to avoid short playlists when seeds get dropped.

The per-artist cap keys on track artist, so it cannot limit a compilation / DJ-mix where every track has a different artist, which is why a per-album cap is needed.

## Test
still testing

## Checklist

**Type of change:**
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [ ] Intel
- [ ] ARM
- [ ] -noavx2
- [x] NVIDIA image

**Tested on media server:**
- [ ] Jellyfin
- [x] Navidrome
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [x] Documentation
- [x] Unit test
- [ ] Integration test

**Other:**

**Related ISSUE:** Closes #525

### DIFF
--- a/app.py
+++ b/app.py
@@ -21,7 +21,7 @@ from flasgger import Swagger, swag_from
 
 # Import configuration
 from config import JELLYFIN_URL, JELLYFIN_USER_ID, JELLYFIN_TOKEN, HEADERS, TEMP_DIR, \
-  REDIS_URL, DATABASE_URL, MAX_DISTANCE, MAX_SONGS_PER_CLUSTER, MAX_SONGS_PER_ARTIST, NUM_RECENT_ALBUMS, \
+  REDIS_URL, DATABASE_URL, MAX_DISTANCE, MAX_SONGS_PER_CLUSTER, MAX_SONGS_PER_ARTIST, MAX_SONGS_PER_ALBUM, NUM_RECENT_ALBUMS, \
   SCORE_WEIGHT_DIVERSITY, SCORE_WEIGHT_SILHOUETTE, SCORE_WEIGHT_DAVIES_BOULDIN, SCORE_WEIGHT_CALINSKI_HARABASZ, \
   SCORE_WEIGHT_PURITY, SCORE_WEIGHT_OTHER_FEATURE_DIVERSITY, SCORE_WEIGHT_OTHER_FEATURE_PURITY, \
   MIN_SONGS_PER_GENRE_FOR_STRATIFICATION, STRATIFIED_SAMPLING_TARGET_PERCENTILE, \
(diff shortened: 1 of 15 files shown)


NeptuneHub:
Just to share that the check on sonic fingerprint was fixed in another PR (My fault, I didn't reocognize that the user opened 2 issue with partially overlap scope).

This PR in my opinion could be viewed as a generalized index for allthe other functionality. Where we check the "max song per artist" we could add also a "max song per album" when you have an album with multiple artist.

rendyhd:
> Just to share that the check on sonic fingerprint was fixed in another PR (My fault, I didn't reocognize that the user opened 2 issue with partially overlap scope).
> 
> This PR in my opinion could be viewed as a generalized index for allthe other functionality. Where we check the "max song per artist" we could add also a "max song per album" when you have an album with multiple artist.

I looked further into this one today. And I think this is no longer needed, is it? I get it for sonic finger print, but is it really needed for the other features. If the perfect compilation matches the request, should we filter it?

NeptuneHub:
You’re right rethinking on the fix made for sonic fingerprint we already avoid to get multiple song on same album.
On other functionality have a top song per artist I think is already good to go. Add top song per album can just overcomplicate the calc. Going to close this.

PR #637, opened 2026-06-13
Feature/swappable sonic backend

PandorasFox:
## AS-IS
Musicnn is the only analysis backend.

## TO-BE
Can swap the analysis backend model with SONIC_BACKEND to mert (95M, cpu/330M, gpu).

Analysis tables are namespaced to the backend used, and unused backend tables are droppable via the cleaning page in the admin panel.

## Test
Tested via switching to the MERT backend and starting a fresh re-analysis of my library. 

## Other useful information
Needed to bump pytorch to 2.6.0 due to a CVE gate. Wrapped up with chunking the audio we infer on after validating everything else works, as that better matches the chunked samples mert was trained on.

## Checklist

**Type of change:**
- [ ] Bug fix
- [X] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [X] Intel
- [ ] ARM
- [ ] -noavx2
- [ ] NVIDIA image

**Tested on media server:**
- [ ] Jellyfin
- [X] Navidrome
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [X] Documentation
- [ ] Unit test
- [ ] Integration test

**Other:**
- [X] CONTRIBUTING.md read and accepted
- [X] Checked performance on a big library (> 150k songs) works without issues

I have ~250K tracks in my music library, and a fresh analysis is projected to take ~11 days on my 14900K (8 workers, 4 threads each) by my napkin math. Doing so with musicnn took something like, idk, 3 or 4 days, but the hit here is expected as mert is just a bigger model.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -413,6 +413,9 @@ RUN set -ux; \
 FROM base AS libraries
 
 ARG BASE_IMAGE
+# Set to 1 to bundle PyTorch + MERT-backend dependencies into the image.
+# Required when SONIC_BACKEND=mert at runtime. Adds ~700 MB on CPU.
+ARG INSTALL_MERT=0
 
 WORKDIR /app
 
@@ -429,6 +432,10 @@ RUN if [[ "$BASE_IMAGE" =~ ^nvidia/cuda: ]]; then \
     else \
         echo "CPU base image: installing all packages together for dependency resolution"; \
         uv pip install --system --no-cache --index-strategy unsafe-best-match -r /app/requirements/cpu.txt -r /app/requirements/common.txt || exit 1; \
+    fi \
+    && if [ "$INSTALL_MERT" = "1" ]; then \
+        echo "INSTALL_MERT=1: adding torch + MERT runtime deps"; \
+        uv pip install --system --no-cache --index-strategy unsafe-best-match -r /app/requirements/mert.txt || exit 1; \
     fi \
     && echo "Verifying psycopg2 installation..." \
     && python3 -c "import psycopg2; print('psycopg2 OK')" \
--- a/app.py
+++ b/app.py
@@ -739,6 +739,17 @@ def listen_for_index_reloads():
               logger.warning(f"SemGrove cache reload failed: {e}")
               sg_success = False
 
+            # Invalidate the per-backend mood centroid cache so the next
+            # request picks up the freshly-rebuilt rows from
+            # mood_centroids_data. Cheap rebuild — handful of JSONB rows.
+            try:
+              import app_voyager as _av
+              _av._mood_centroids_loaded = False
+              _av._MOOD_CENTROIDS_DATA.clear()
+              _av._MOOD_CENTROIDS_META.clear()
+            except Exception as e:
+              logger.warning(f"Mood centroid cache invalidation failed: {e}")
+
             logger.info(f"In-memory reload complete: Voyager ✓, Artist ✓, Maps ✓, CLAP {'✓' if clap_success else '✗'}, Lyrics {'✓' if lyrics_success else '✗'}, SemGrove {'✓' if sg_success else '✗'}")
           except Exception as e:
             logger.error(f"Error reloading indexes/maps from background listener: {e}", exc_info=True)
--- a/app_analysis.py
+++ b/app_analysis.py
@@ -176,42 +176,37 @@ def start_cleaning_endpoint():
 
 @analysis_bp.route('/api/cleaning/sonic_state', methods=['GET'])
 def sonic_state_endpoint():
-    """Report current vs stored sonic-backend embedding state.
+    """Report per-backend embedding + Voyager state for the Cleaning UI.
 
-    Used by the Cleaning UI to surface a "stale audio embeddings" panel
-    after a ``SONIC_BACKEND`` change. The body is exactly whatever
-    ``tasks.cleaning.inspect_sonic_state`` returns:
-    ``current_backend``, ``current_dim``, ``embedding_row_count``,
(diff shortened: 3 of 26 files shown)


NeptuneHub:
Hi and thanks for this contribution. Can you explain me a bit better ?

- what you think don’t work good on Musicnn model ?
- which better result did you reached with this new model? Can you share your hands on? Possible even some example of old vs new playlist;
- For switchable what you mean? A new user wake up in the morning and download audiomuse-ai, how the switch is supposed to work for him? And how for legacy users ?
- can you link the repository and the license of the model that you used? Is compatible with AGPLv3?
- the new model how and when is downloaded? It is on third party repository ?
- You think about switch Musicnn that is probably one of the smaller with better result related to its size to mert, which other model did you research about before arriving to mert? Can you share the result of your analysis ?
- Can this model work also on ARM and noavx2 Intel machine on top of Intel ? What about the use of it on Nvidia GPU?
- do you write that you’re using 8 workers on the same machine/cpu, why not only one ? Did you noticed increased speed?
- What about the use of RAM ? During the analysis which is the avarage use of ram? And which  is the peach usage?
- Can you also share the usage of ram in idle with the full index computer on your library with Musicnn and with mert ?
- Switching the model all the functionality that use Musicnn are able to use this new model ? Did you tested ?

Thank you.

PandorasFox:
> Hi and thanks for this contribution. Can you explain me a bit better ?

Sure :)


>     * what you think don’t work good on Musicnn model ?

Musicnn is notably a bit limited/inaccurate at genre classification due to its limited training pool. I mostly use audiomuse with the navidrome plugin for song similarity, but I find that in my ~250k track music library, I get a lot of crossover between genres due to mis-classification (e.g. jazz miscategorized as drums n bass and vice versa). 

Broadly, mert outperforms musicnn on genre & similarity scoring, so I'm happy with the slower inference/analysis throughput in exchange for better inference quality in this space.


>     * which better result did you reached with this new model? Can you share your hands on? Possible even some example of old vs new playlist;
> 
>     * For switchable what you mean? A new user wake up in the morning and download audiomuse-ai, how the switch is supposed to work for him? And how for legacy users ?

By default, musicnn is used. If a user wishes to use, say, mert 330M w/ their gpu, they can set the sonic_backend env var, the new model will be pulled in on rebuild, and the inferences / analysis will end up in new tables, with more vectors on the similarity analysis, namespaces by the model used.


>     * can you link the repository and the license of the model that you used? Is compatible with AGPLv3?

https://huggingface.co/m-a-p/MERT-v1-95M - creative Commons non-commercial. i am not a licensing expert but i believe this is fine?


>     * the new model how and when is downloaded? It is on third party repository ?

Right now, the image itself needs to be built with a flag to enable mert being pulled in and baked into the image. not really ideal (but works quite well for testing); producing a prebuilt audiomuse-mert image or something via CI might be preferable.

>     * You think about switch Musicnn that is probably one of the smaller with better result related to its size to mert, which other model did you research about before arriving to mert? Can you share the result of your analysis ?

I also looked at jukebox and musicFM, but from my review of the whitepapers, MERT has the overall best accuracy on similarity/genre scoring, so I opted for it.

>     * Can this model work also on ARM and noavx2 Intel machine on top of Intel ? What about the use of it on Nvidia GPU?

Unsure about arm & noavx2. Mert is considerably weightier than musicnn and probably wouldn't perform well on noavx2. the 330M large mert model works great on nvidia, and the standard mert v1-95M works great on x86 w/ avx2 

>     * do you write that you’re using 8 workers on the same machine/cpu, why not only one ? Did you noticed increased speed?

I'm running 8 workers with 4 pytorch threads each on a 14900k. More than that and I run into cpu contention, but this is the sweet spot for the best analysis throughput on my homelab

>     * What about the use of RAM ? During the analysis which is the avarage use of ram? And which  is the peach usage?

Each worker uses between 750MiB and 1.4GiB of ram at runtime, with just shy of 8GiB _total_ (not each) in use by the 8 workers at any given moment. Peak could be up to ~12GiB.

>     * Can you also share the usage of ram in idle with the full index computer on your library with Musicnn and with mert ?

If you mean with the workers idle - I can do this comparison ~next week once my full analysis run has completed 😅 a full analysis pass of my very large library took 4 or 5 days with musicnn, and is estimated to take about 10 days with mert on the same hardware. Idle usage probably just sits at ~700MiB per worker since the model itself is about 700MiB?

>     * Switching the model all the functionality that use Musicnn are able to use this new model ? Did you tested ?

It should be, yeah. The centroids bits downstream of the full analysis should also be namespaced per model and work fine once my analysis pass is done, but it'll be a bit before I have enough infefence done to validate & compare quality.  I'll be testing that properly in a week or so.

> Thank you.

Thank you for all the work on audiomuse :)

NeptuneHub:
CC-BY-NC is not aligned at all with AGPLv3.

AGPLv3 enalbe the use for commercial purpose, CC-BY-NC not.

I'm going to close this because violate the license term. I'll also improve the contribution guide to add a clear step on the license to be more visible.

Thanks for your effort but we can't violate a license.

PandorasFox:
> CC-BY-NC is not aligned at all with AGPLv3.
> 
> AGPLv3 enalbe the use for commercial purpose, CC-BY-NC not.
> 
> I'm going to close this because violate the license term. I'll also improve the contribution guide to add a clear step on the license to be more visible.
> 
> Thanks for your effort but we can't violate a license.

understandable! https://github.com/PandoraMedia/music-audio-representations would be my other look for a larger gpl model, but i've zero experience with it or its quality.

PR #688, opened 2026-06-23
Update native Linux instructions documentation in README.md

life1301:
corrected native Linux installation instructions to reflect current release naming scheme

**Type of change:**
- [ ] Bug fix
- [ ] New feature
- [X] Documentation
- [ ] Refactor
- [ ] Breaking change

**Updated:**
- [X] Documentation
- [ ] Unit test
- [ ] Integration test


NeptuneHub:
Have you checked the Gemini review? are they applicable?

life1301:
It is appropriate. I'm going to work on headless instructions this week if you want to hold off on this.

life1301:
added instructions required for Trixie and suggested for Bookworm. As stated in the new instructions, testing was completed with respective versions of standard Debian LXC templates on Proxmox. I am unable to test on bare metal at this time.

NeptuneHub (changes requested):
Please remove all this part:
```
Required prerequisite steps necessary on Debian 13 (Trixie) headless/minimal installs:
Tested on Proxmox LXC with standard Trixie template. Feedback on other Trixie deployments is welcome.
Do not run these commands as root!
Setup a user and run all commands as this user.
echo 'export XDG_RUNTIME_DIR="/run/user/$(id -u)"' >> ~/.bashrc
sudo apt update
sudo apt install xdg-utils libgomp1 libpam-systemd -y
sudo loginctl enable-linger $USER
Restart, log back in as the same user, and follow the remaining instructions.
Suggested prerequisite steps for Debian 12 (Bookworm) headless/minimal installs:
Tested on Proxmox LXC with standard Bookworm template.
sudo apt install xdg-utils libgomp1 (prevents dpkg failing on dependency checks which requires sudo apt --fix-broken install to fix.)
```
All the native version have the goal to be easy to install, for non tech user that install AudioMuse-AI on their desktop. So all this part only add complexity that instead we want to avoid.

Also I noticed a em dash, please remove.
Thanks.

life1301:
While I do understand the utility of simplicity, I would argue native deb packages are super useful to users like myself who containerize applications such as this. Having to use Docker in a VM or LXC nests virtualization which creates unnecessary overhead. Running in an LXC easily isolates environments and allows us to passthrough a shared GPU vs VM with desktop would isolate the passthrough GPU to that VM only.

I did spin up a Trixie VM with GNOME and did not have problems so these issues appear to be isolated to headless installs only. Might a more elegant solution be achieved by adding these checks to the native Linux build postinstall.sh to only fire on headless installs and libpam-systemd dependency to nfpm.yaml.in? This would qualify your goal of low complexity to the end user. If you are open to this I can work that route instead.

NeptuneHub (approved):

PR #738, opened 2026-07-09
fix: require current password for user management

ShiroKSH:
﻿## AS-IS
Admins could create additional users without confirming their own password. Admin resets for admin accounts also skipped current-password confirmation, while only non-admin self-service password changes required it.

## TO-BE
Creating a user requires the current authenticated admin password. Password changes require current-password confirmation for self-service changes and when the target account is an admin, while admin resets for regular user accounts keep the existing flow.

## Test
- `uv run --python 3.12 --with pytest --with flask --with PyJWT --with argon2-cffi --with psycopg2-binary --with redis --with rq --with pyyaml --with requests --with numpy --with rapidfuzz pytest test/unit/test_app_auth.py`
- `uv run --with ruff ruff check app_auth.py test/unit/test_app_auth.py`
- `node --check static/users.js`
- `git diff --check`

## Other useful information
Split from closed #735 and limited to user-management password confirmation.

## Checklist
**Type of change:**
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [ ] Intel
- [ ] ARM
- [ ] -noavx2
- [ ] NVIDIA image

**Tested on media server:**
- [ ] Navidrome
- [ ] Jellyfin
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [ ] Documentation
- [x] Unit test
- [ ] Integration test

**Other:**

**Related ISSUE:** N/A

### DIFF
--- a/app_auth.py
+++ b/app_auth.py
@@ -318,6 +318,21 @@ def verify_additional_user(username, password):
     return _normalize_role(role) or USER_ROLE_USER
 
 
+def _current_password_error(data):
+    current_username = getattr(g, 'auth_user', None)
+    current_password = data.get('current_password') if isinstance(data, dict) else None
+    if (
+        not isinstance(current_username, str)
+        or not current_username
+        or not isinstance(current_password, str)
+        or not current_password
+    ):
+        return "Current password is required."
+    if verify_additional_user(current_username, current_password) is None:
+        return "Current password is incorrect."
+    return None
+
+
 def upsert_admin_user(username, password):
     """Create an admin row, or update the password and force admin role when
     the username already exists. Returns ``(ok, error_message)``.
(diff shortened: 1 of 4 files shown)


NeptuneHub (changes requested):
Hi and thanks for the PR. Please the current password field, when is requested for activity on OTHER users (so not yourself) need to be clear that is your password to confirm the activity so:
- Instead of current password label put something like "Insert your admin password to confirm the operation"
- put it in the bottom, separated maybe with some space from the information of the user that you are creating/modifing

NeptuneHub:
Thanks for taking the time to put this together and for the suggestion. I think the underlying idea is worthwhile, but after reviewing the changes, I don't think this implementation is the right fit.

At this point, I think it will be more efficient for me to implement the change directly rather than iterate further on this PR, so I'm going to close it.

Thanks again for the contribution and the idea.

PR #739, opened 2026-07-09
fix: validate backup restore uploads before locking

ShiroKSH:
﻿## AS-IS
Empty backup restore uploads and malformed chunk metadata could reach the restore-lock path before being rejected later or failing during processing.

## TO-BE
Restore uploads reject empty files and require `chunk_num` plus `total_chunks` to be submitted together before the restore lock is acquired. The backup page also blocks zero-byte restore files before upload.

## Test
- `uv run --python 3.12 --with pytest --with flask --with PyJWT --with argon2-cffi --with psycopg2-binary --with redis --with rq --with pyyaml --with requests --with numpy --with rapidfuzz pytest test/unit/test_app_backup_restore.py`
- `uv run --with ruff ruff check app_backup.py test/unit/test_app_backup_restore.py`
- `git diff --check`

## Other useful information
Split from closed #735 and limited to backup restore upload validation.

## Checklist
**Type of change:**
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [ ] Intel
- [ ] ARM
- [ ] -noavx2
- [ ] NVIDIA image

**Tested on media server:**
- [ ] Navidrome
- [ ] Jellyfin
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [ ] Documentation
- [x] Unit test
- [ ] Integration test

**Other:**

**Related ISSUE:** N/A

### DIFF
--- a/app_backup.py
+++ b/app_backup.py
@@ -85,6 +85,21 @@ def _restore_lock_held():
         return True
 
 
+def _uploaded_file_is_empty(uploaded):
+    stream = getattr(uploaded, 'stream', None)
+    if stream is None:
+        return False
+    try:
+        pos = stream.tell()
+        stream.seek(0, os.SEEK_END)
+        size = stream.tell()
+        stream.seek(pos)
+        return size == 0
+    except (AttributeError, OSError):
+        content_length = getattr(uploaded, 'content_length', None)
+        return content_length == 0
+
+
 def _pg_env():
     """Return a copy of os.environ with PGPASSWORD set."""
     env = os.environ.copy()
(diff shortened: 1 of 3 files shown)


NeptuneHub:
Closing, not merging. The underlying concern is valid and I'll track it in a separate issue, but the PR is not at a mergeable standard.

Beyond the process issue, the fix itself is half-baked — it only handles the zero-byte case:

- A truncated or corrupted non-empty dump passes your check and still reaches the restore path, which is the same data-loss scenario you claim to fix. You check size, never content (no pg_dump header/format validation).
- The bot already flagged that `tell()`/`seek()` can raise `ValueError` (closed stream), which your except clause doesn't catch → 500. Still unaddressed.
- A non-seekable upload without Content-Length gets falsely rejected as empty by your fallback.
- If an empty chunk N>1 is rejected, the restore lock acquired on chunk 1 stays held for up to 1 hour and partial temp files are left behind — so "validate before locking" is only true for the first chunk.

As stated on #735: AI-assisted PRs are only accepted when the submitter has personally reviewed and tested the change on a real deployment. Leaving bot feedback unaddressed and not covering the obvious adjacent cases tells me that didn't happen here. Please don't open further PRs that don't meet that bar.

PR #741, opened 2026-07-09
fix: render dynamic UI data with text nodes

ShiroKSH:
﻿## AS-IS
Several UI paths rendered API or library metadata by interpolating it into HTML strings. Some error and selection messages also used inline HTML event handlers or string-built table rows.

## TO-BE
High-risk UI paths now render dynamic names, messages, table cells, autocomplete rows, map legend entries, playlist rows, and selected labels with DOM nodes and `textContent` instead of interpolated HTML.

## Test
- `node --check static/script.js`
- `node --check static/sunburst.js`
- `git diff --cached --check`
- `rg -n 'onclick=' static/script.js static/sunburst.js templates/alchemy.html templates/clap_search.html templates/dashboard.html templates/lyrics_search.html templates/map.html templates/path.html templates/similarity.html templates/waveform.html` returned no matches.
- Reviewed scoped `innerHTML` template-literal grep; remaining matches are static layout snippets or existing generated markup that does not interpolate API text directly.

## Other useful information
Split from closed #735 and limited to safe rendering in static/template UI code.

## Checklist
**Type of change:**
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [ ] Intel
- [ ] ARM
- [ ] -noavx2
- [ ] NVIDIA image

**Tested on media server:**
- [ ] Navidrome
- [ ] Jellyfin
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [ ] Documentation
- [ ] Unit test
- [ ] Integration test

**Other:**

**Related ISSUE:** N/A

### DIFF
--- a/static/script.js
+++ b/static/script.js
@@ -1,10 +1,3 @@
-// Escape user/library-derived strings before inserting them into innerHTML.
-// Playlist names and track title/author come from media-server metadata, which
-// is attacker-influenceable, so they must be HTML-escaped to prevent stored XSS.
-const escapeHtml = (str) => String(str ?? '').replace(/[&<>"']/g, c => (
-    { '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;', "'": '&#39;' }[c]
-));
-
 // --- DOM Element References ---
 const mainContent = document.getElementById('main-content');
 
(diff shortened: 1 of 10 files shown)


NeptuneHub:
This PR change 10 files to address the case that AudioMuse-AI selfhosted from the user, get malicius input from the Music Server selfhosted from the same users. 

This scenario could maybe happen if the users put song with malicius metadata in his music server and the music server return them totally unescaped and then AudioMuse-AI inject it.

So you're describing an attack scenario that should:
- bypass the Music Server control
- be able to attack AudioMuse-AI.

To me the risk is higher to merge your untested PR than this possible but maybe not probable attack scenario.

I'm not going to merge this.

PR #742, opened 2026-07-09
fix: revalidate JWT session users

ShiroKSH:
﻿## AS-IS
JWT cookies trusted the role claim already stored in the token. If a user's role changed or the user row was removed after login, an existing token could keep using stale authorization until it expired.

## TO-BE
JWT cookies with a `sub` now re-read the user's role from `audiomuse_users` on each authenticated request and reject sessions for deleted users. Legacy tokens without `sub` keep the existing admin-compatible fallback.

## Test
- `uv run --python 3.12 --with pytest --with flask --with PyJWT --with argon2-cffi --with psycopg2-binary --with redis --with rq --with pyyaml --with requests --with numpy --with rapidfuzz pytest test/unit/test_app_auth.py`
- `uv run --with ruff ruff check app_auth.py test/unit/test_app_auth.py`
- `git diff --check`

## Other useful information
Split from closed #735 and limited to JWT session user/role validation.

## Checklist
**Type of change:**
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [ ] Intel
- [ ] ARM
- [ ] -noavx2
- [ ] NVIDIA image

**Tested on media server:**
- [ ] Navidrome
- [ ] Jellyfin
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [ ] Documentation
- [x] Unit test
- [ ] Integration test

**Other:**

**Related ISSUE:** N/A

### DIFF
--- a/app_auth.py
+++ b/app_auth.py
@@ -318,6 +318,31 @@ def verify_additional_user(username, password):
     return _normalize_role(role) or USER_ROLE_USER
 
 
+def _session_role_for_user(username):
+    if not isinstance(username, str) or not username:
+        return None
+    db = None
+    try:
+        db = _get_db()
+        with db.cursor() as cur:
+            cur.execute(
+                "SELECT role FROM audiomuse_users WHERE username = %s",
+                (username,),
+            )
+            row = cur.fetchone()
+    except Exception:
+        logger.exception("Failed to validate session user %r", username)
+        if db is not None:
+            try:
+                db.rollback()
+            except Exception:
+                pass
+        return None
+    if not row:
+        return None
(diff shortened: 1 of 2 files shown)


NeptuneHub:
Thanks for taking the time to put this together and for the suggestion. I think the underlying idea is worthwhile, but after reviewing the changes, I don't think this implementation is the right fit.

At this point, I think it will be more efficient for me to implement the change directly rather than iterate further on this PR, so I'm going to close it.

Thanks again for the contribution and the idea.

PR #791, opened 2026-07-24
Chunk MusiCNN embedding inference to keep peak memory flat

deluan:
## Problem

`_run_musicnn_models` feeds **every spectrogram patch of a track through the embedding model in a single ONNX batch**. The convolution activations for that batch scale linearly with track length: a ~10-minute track (99 patches) peaks at **~4.7 GB RSS** on CPU. That blows past typical worker memory caps (containers commonly run with 1.5–4 GB limits), and is a likely contributor to the OOM kills / signal-9 crashes the codebase already works around (restart-after-N-jobs, `malloc_trim`, `handle_onnx_memory_error`, the GPU-OOM→CPU fallback).

## Fix

Run the embedding inference in small chunks of patches instead of one whole-track batch. Chunking happens around `run_inference_with_oom_fallback`, so the existing GPU-OOM→CPU fallback and session-swap bookkeeping still work per chunk.

New setting **`MUSICNN_BATCH_SIZE`** (env var, default `8`, documented in `docs/PARAMETERS.md`). `0` restores the previous whole-track behavior. GPU users who prefer larger batches for throughput can raise it.

Only the embedding model is chunked; the prediction model's input is the (N, 200) embedding matrix, which is tiny.

## Measurements

23 real tracks (12–99 patches each, 1257 patches total), CPU EP, Apple M4 (arm64), models from `v5.0.0-model`, calling `_run_musicnn_models` from this branch:

| | wall time | peak RSS |
|---|---|---|
| `MUSICNN_BATCH_SIZE=0` (previous behavior) | 9.6 s | **4691 MB** |
| `MUSICNN_BATCH_SIZE=8` (default) | 8.7 s | **391 MB** |

**Output parity:** pooled embeddings match the full-batch output at cosine ≥ 0.999999941 (ONNX Runtime run-to-run reduction noise), final mood scores are bit-identical, and the 200-bit simhash canonical IDs are unchanged on all 23 tracks — so existing libraries need no re-analysis and mixed old/new worker fleets stay consistent.

## How it was tested

- **Unit tests added** (`test/unit/test_musicnn_chunking.py`): embedding calls never exceed `MUSICNN_BATCH_SIZE`, every patch is processed exactly once and in order, and chunked pooled outputs are identical to the whole-track batch. Full `pytest test/unit/` passes (2575 passed) with this change; `flake8 --select=E9,F,W605,E711,E712,E713,E714,E722,E401` is clean on the touched files.
- **Functional parity:** the 23-track comparison above ran the real `_run_musicnn_models` from this branch against the previous implementation's outputs (same models, same patches), measuring peak RSS via `ru_maxrss` in separate processes. Analysis was exercised with audio decoded from a Navidrome library (MP3 + FLAC).
- To replicate: run analysis on any album with `MUSICNN_BATCH_SIZE=0` vs unset while watching worker RSS; long tracks (8+ minutes) show the difference most clearly.

### DIFF
--- a/config.py
+++ b/config.py
@@ -622,6 +622,13 @@ CLAP_PYTHON_MULTITHREADS = os.environ.get("CLAP_PYTHON_MULTITHREADS", "False").l
 #   Cons: May see gradual VRAM growth on some systems
 PER_SONG_MODEL_RELOAD = os.environ.get("PER_SONG_MODEL_RELOAD", "true").lower() == "true"
 
+# Maximum number of spectrogram patches sent through the MusiCNN embedding
+# model in a single ONNX inference call. Running a whole track as one batch
+# (the previous behavior) makes the convolution activations peak at several
+# GB of RAM/VRAM on long tracks; small batches keep peak memory flat with no
+# measurable speed cost. Set to 0 to disable chunking (one whole-track batch).
+MUSICNN_BATCH_SIZE = int(os.environ.get("MUSICNN_BATCH_SIZE", 8))
+
 # Category weights for CLAP query generation (affects random query sampling probabilities)
 # Higher weights favor categories where CLAP excels (Genre, Instrumentation)
 # Format: JSON string with category names as keys and float weights as values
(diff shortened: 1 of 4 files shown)


NeptuneHub:
OK, looking at integraiton test seems that the embbeding procuded are the same, and I like the idea of less ram consumed, so I'm going ot merge.

NeptuneHub (approved):

Issue #805, opened 2026-07-26
[BUG] Failed/abandoned RQ jobs leave task_status in PROGRESS forever and block new tasks


rendyhd:
**Describe the bug**

AudioMuse-AI v3.0.5 can leave a main analysis permanently displayed as active after its worker is interrupted.
In a split server/worker deployment, Docker Desktop restarted the worker while analysis was performing Chromaprint backfill. RQ correctly moved the job to FailedJobRegistry with:
```
AbandonedJobError
status: failed
retries_left: 0
```
However, PostgreSQL remained:
```
task_type: main_analysis
status: PROGRESS
progress: 99
end_time: NULL
```
Both RQ queues were empty and the replacement workers were idle, but /api/active_tasks continued returning the PostgreSQL row. The UI therefore remained at 99%, and starting another analysis was blocked with HTTP 409.

**Root cause:**
reap_orphaned_tasks() in tasks/multiserver_sync.py only marks a task failed when Job.fetch() raises NoSuchJobError. An abandoned job still exists in Redis with status failed, so it is incorrectly treated as live:
```
try:
    Job.fetch(task_id, connection=redis_conn)
except NoSuchJobError:
    reaped.append(task_id)
```

**Expected behavior**

The janitor should treat only these RQ statuses as active:
queued, started, deferred, scheduled
If a non-terminal PostgreSQL task has an RQ status such as failed, stopped, canceled, or finished, it should be reconciled to an appropriate terminal database status.

**Environment (please complete the following information):**

 - OS: Docker/Windows flask/worker
 - AudioMuse-AI Version: v305
 - Navidrome/Jellyfin Version: Navidrome

NeptuneHub:
Going to close because PR is merged. Let me know in case this issue come back again and feel free to raise new pr. thanks.

PR #806, opened 2026-07-26
Fix stale task status after RQ job termination

rendyhd:
PR for #805


NeptuneHub:
The fix looks good from a code inspection, just some point to give a look:
- **Centralized status function:** Four places now decide if a job is alive: `_sweep_job_state`, your `_RQ_TERMINAL_STATUSES`, `_rq_job_still_pending` (misses `deferred`), the clustering monitor (misses `stopped`). Nothing broken today, but that is how #805 started. Can you think about a centralized helper?

- **Additional unit test:** A Redis error is not a dead job. Your code does this right already, but nothing locks it in. Just one more unit test in `test/unit/test_multi_server.py` (no app code change): `Job.fetch` raises `ConnectionError`, assert 0 rows updated. Otherwise someone later merges this probe with `_sweep_job_state`, which treats any error as "job gone", and one Redis timeout marks every running task as FAILURE.

- **Additional integration test:** The `AND status NOT IN (SUCCESS, FAILURE, REVOKED)` guard is the best part of the PR, since a task writes SUCCESS before RQ marks the job finished. But the tests only check SQL parameters on a mock, which cannot prove a row survived. As an example, `test/integration/test_task_status_integration.py` already spins up a real Postgres with a real `task_status` table, so the two cases fit naturally in there: a PROGRESS row with a failed job becomes FAILURE, and a row that turns SUCCESS mid-pass stays SUCCESS. It would also fix the 0% coverage from SonarCloud.

NeptuneHub:
Hi @rendyhd I passed the afternoon in doing test and fix, in details:
- we centralized the RQ job status logic in the new rq_job_state.py, so every part of the app checks job state the same way
- the analysis chromaprint step kept running even if cancelled, and was requeued after a restart: fixed it by removing the job's retry budget on cancel and checking for cancel before writing progress
- clustering was requeued after a crash but then finished with FAILURE: fixed it by writing FAILURE only when no retries are left

Also: cron jobs now retry like UI jobs, safer janitor DB updates, fewer Redis calls, new tests.

I did several test by restarting worker/flask container while:
-  Analyis running in initial enque stage, during the analysis, during the index creation and during chromaprting creation => It re-started correctly;
- Analysis CANCELLED in different stage => it remain cancelled;
- Clustering running => it restarted and completed;

I ask if you can give some other try and confirm before I proceed with merge. Thanks!

PR #837, opened 2026-08-03
feat(media-servers): add K7 as a media-server provider

kaybi-gh:
Wire [K7](https://github.com/kaybi-gh/K7) into the dispatcher, setup wizard, multi-server admin, and provider migration so it can be used like the existing backends.

**Note on K7:** K7 has just reached its **1.0.0** release and has not been publicly shared with the self-hosted / media-server community yet, so it will be unfamiliar to most reviewers. I fully understand if you prefer not to merge a provider for a project that is not widely known or adopted yet. I will still be happy to revisit later once K7 is public and has more real-world use.

## AS-IS
does not exist

## TO-BE
AudioMuse-AI supports **K7** as a first-class media-server provider (`MEDIASERVER_TYPE=k7`), alongside Jellyfin / Navidrome / Emby / Lyrion / Plex.

Users can:
- pick **K7** in the Setup Wizard and Music Servers admin (URL + API key)
- analyze libraries, download tracks, manage playlists, and use Instant Playlist against K7
- use K7 as a **provider-migration target** (and as a source for path-refresh when it is the default server)

Config keys: `K7_URL`, `K7_API_KEY` (default HTTP port **7080**).

## Test
Tested locally against a running K7 **1.0.0** instance and AudioMuse-AI from this branch:
1. Setup Wizard -> Media server type **K7** -> URL `http://<host>:7080` + API key -> Test connection -> Save
2. Music Servers admin: add/set K7 as default
3. Library listing / analysis start / clustering
4. Instant Playlist creation (playlist appears on K7 with `_instant` suffix)
5. Optional: Provider Migration -> target **K7** -> Test connection

Unit: `test/unit/test_mediaserver_k7.py` (playlist id return, `_instant` naming, case-insensitive library filter, zero-play top-tracks fallback).

Not run on NVIDIA / ARM / `-noavx2` images for this PR; not a full regression pass on Navidrome/Jellyfin/Emby/Lyrion/Plex.

## Other useful information
- Implementation lives in `tasks/mediaserver/k7.py` and follows the same lazy-dispatch / UI wiring pattern as Plex.
- Downloads: direct-stream first, then `downloads/prepare` fallback (ApiKey clients).
- `get_top_played_songs` returns tracks even when play counts are 0 (fresh libraries), so sonic fingerprint / probes are not starved.
- Playlist item adds are one HTTP call per track (K7 has no batch add endpoint yet).
- Path-based multi-server matching depends on K7 exposing file paths in list payloads; otherwise fingerprint/metadata matching applies.
- Docs/README updated (badge lists **K7 1.0.0**). Local-only `docker-compose.local.yaml` is not included.
- PR checklist template also lists Plex and K7 under "Tested on media server".

## Checklist

**Type of change:**
- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [x] Intel
- [ ] ARM
- [ ] -noavx2
- [x] NVIDIA image

**Tested on media server:**
- [ ] Navidrome
- [ ] Jellyfin
- [ ] Emby
- [ ] Lyrion
- [ ] Plex
- [x] K7

**Updated:**
- [x] Documentation
- [x] Unit test
- [ ] Integration test

**Other:**
- [x] Checked performance on a library (> 23k songs) works without issues


NeptuneHub:
Thanks but I'm not going to merge this for a Client at the moment is not used.

I higly reccomand to implement Open Subsonic API as documented here:
- https://opensubsonic.netlify.app/

This not only will give you the AudioMuse-AI support, but also the support for multiple music player out of there.

kaybi-gh:
Understandable, thank you for taking the time to answer!

I had considered Subsonic earlier but held off because the legacy auth model felt too weak for something I want to ship. Since OpenSubsonic supports API keys, do you also support/recommend that path on the AudioMuse side? If so, that would make an OpenSubsonic implementation much more reasonable from a security standpoint.

NeptuneHub:
If you implement the API KEY authentication as documented in the Open Subsonic standard here:
- https://opensubsonic.netlify.app/docs/extensions/apikeyauth/
- https://opensubsonic.netlify.app/docs/api-reference/

Then you can eventually submitt a small PR against audiomuse-ai to change navidrome.py to supported the API token login instead. Also Setup Wizard and Provider Migration will need to be changed but at least we are talking about only 1 parameter that could be reused in future also from other music server.

PR #845, opened 2026-08-08
feat(navidrome): add OpenSubsonic apiKey authentication

kaybi-gh:
Follow-up to the discussion on #837: instead of adding a custom media-server provider, this PR teaches the existing Navidrome/OpenSubsonic path to use [OpenSubsonic API key auth](https://opensubsonic.netlify.app/docs/extensions/apikeyauth/) as suggested.

## AS-IS
Navidrome / Subsonic connections only support username + password (`u` / `p=enc:...`). Servers that expose OpenSubsonic `apiKeyAuthentication` cannot be configured without a password.

## TO-BE
Navidrome (and any OpenSubsonic-compatible server selected as Navidrome) can authenticate with either:
- username + password (unchanged) or
- an OpenSubsonic API key (`apiKey=`, no `u` parameter)

Capture d&#39;écran 2026-08-08 101140

Capture d&#39;écran 2026-08-08 101206

Setup Wizard, Music Servers admin, and Provider Migration expose an explicit auth-method choice (password vs API key). One reusable credential field: `NAVIDROME_API_KEY` / `api_key`.

Existing password-only installs keep working and open on the password option by default.

## Test
1. Setup Wizard -> Media server type **Navidrome** -> choose **Username + password** -> Test connection -> Save
2. Same wizard -> choose **OpenSubsonic API key** -> URL + API key only -> Test connection -> Save
3. Confirm inactive credentials are cleared when switching method and saving
4. Provider Migration -> target Navidrome -> same auth-method radios -> Test connection
5. Music Servers admin -> add/edit a Navidrome server with each auth method

Also verified end-to-end against a K7 server as an OpenSubsonic server that implements `apiKeyAuthentication` (library probe, analysis resume with existing item IDs).

Unit:
- `test/unit/test_mediaserver.py` (apiKey preferred, no `u`/`p`; missing creds XOR)
- `test/unit/test_setup_manager.py` (apiKey-only and password-only validity)
- `test/unit/test_multi_server.py` (`api_key` masking / clear on merge)

Not a full regression pass on Jellyfin / Emby / Lyrion / Plex. Not tested on ARM / `-noavx2`.

## Other useful information
- Scope is intentionally small: no new provider type (we could have duplicated code to have a brand new and specific Subsonic entry), only Navidrome/OpenSubsonic auth + wizard/migration UI + docs/tests.
- Spec compliance: when API key mode is used, clients must not send `u` (avoids OpenSubsonic error 43).
- `navidrome_auth_mode` on setup save/test:
  - `apikey` -> clear stored username/password
  - `password` -> clear stored API key
  - missing / empty / unknown -> **no-op** (do not rewrite credentials). Older clients and non-UI API callers keep working; runtime still uses `api_key` when present, otherwise username/password.
- Docs: `docs/PARAMETERS.md`, `docs/NAVIDROME.md`.

## Checklist

**Type of change:**
- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Refactor
- [ ] Breaking change

**Tested on architecture:**
- [x] Intel
- [ ] ARM
- [ ] -noavx2
- [x] NVIDIA image

**Tested on media server:**
- [x] Navidrome (K7 in fact, but I also did spin up a navidrome instance to ensure there was no regression)
- [ ] Jellyfin
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [x] Documentation
- [x] Unit test
- [ ] Integration test

**Other:**
- [ ] Checked performance on a big library (> 23k songs) works without issues
image

**Related ISSUE:** Related to #837


NeptuneHub:
Thanks for the PR. Please address the merge conflict from main and the SonarCloud code analysis and give a final check to look that everything still work. After that I'll do a round of test also on my side and eventually merge it.

Looking forward to support K7 with this PR!

kaybi-gh:
Thank you for answering this fast!
I did address the merge conflict issues and treated SonarCloud code analysis. 👍

> Looking forward to support K7 with this PR!

Not as much as I do 😁 
I've been using your project on my K7 instance for a while now and it's really super super cool. I will need to teach some of my family and friends english to make the best use of it though.. 😂 

Now you have to do the same for movies, series and podcasts then make it a public online db to get suggestions from outer space 😏

NeptuneHub (changes requested):
Hi,
I asked to Claude Sonnet, GPT 5.6 and Deepseekv4 flash to review it and  I found two main problems before this PR can be merged.

The OpenSubsonic API key implementation itself looks correct, but the new authentication logic does not fully handle multi-server setups. Credentials from the default Navidrome server can be used against another Navidrome server. There is also an issue in the Setup Wizard library discovery, and the Unit Tests workflow is currently failing.

Please implement it and test configuring 2 server: Navidrome with user&password and yours with API Key (frist navidrome default, then try yours default). Check if everything still work.

Below a recap from the 3 AI, off course get it with a grain of salt.

## 1. Critical: wrong credentials can be sent to another Navidrome server

`get_navidrome_auth_params()` falls back to the global `NAVIDROME_API_KEY` when the selected server has no API key, even if valid username/password credentials were explicitly provided for that server.

Example:

* Default Navidrome uses `API_KEY_A`
* Secondary Navidrome uses username/password
* A request is made to the secondary server

The request can use the secondary server URL together with `API_KEY_A`.

So the secondary username/password is ignored, authentication fails, and more importantly the default server API key can be leaked to another host.

This affects normal multi-server operations and Provider Migration when the default/source uses an API key and the target uses username/password.

The fallback pattern existed before, but this PR makes an empty `api_key` or empty `user/password` a valid configuration, which exposes the problem.

Credential resolution should be atomic: once server-specific credentials are selected, only those credentials should be used. Global config should only be used when no explicit or bound server credentials exist.

## 2. Setup Wizard library listing

Test Connection correctly receives `navidrome_auth_mode` and clears the inactive authentication method.

The following library-list request does not receive or apply `navidrome_auth_mode`.

So when switching from API key to username/password, Test Connection can succeed with the new password, while library discovery can still fall back to the previously saved API key.

The library-list endpoint should apply the auth mode in the same way as Test Connection.

## 3. Spec compliance and other providers

The OpenSubsonic implementation itself looks correct:

* `apiKey` is sent without `u`/`p`
* auth errors `42`, `43` and `44` are handled
* `apiKey` is redacted from logs
* old single-server username/password auth still works
* Jellyfin, Emby, Plex and Lyrion are not affected by these changes

One small note: some secret handling for `NAVIDROME_API_KEY` depends on the `_API_KEY` field-name suffix rather than explicitly listing the field. It works today, but it is slightly implicit.

## 4. Missing tests / current CI

The new tests cover the normal single-server cases but not the important mixed-auth multi-server case.

I would add tests for:

* default Navidrome with API key + secondary Navidrome with password
* explicit password credentials while a global API key exists
* Provider Migration from API-key Navidrome to password Navidrome
* switching auth mode in Setup Wizard and then loading libraries

Also, the Integration Tests are passing but the Unit Tests workflow is currently failing on the PR head.

kaybi-gh:
Thank you for the review! The multi-server leak was a real one, good catch.

I pushed a fix:
- Auth is atomic now, if a bound/explicit server is in play, we stick to its creds only. No more grabbing the default server’s API key by accident.
- Library listing in the Setup Wizard now gets navidrome_auth_mode the same way Test Connection does (that mismatch was silly..).
- Unit tests fixed + added mixed-auth multi-server cases. CI should pass now!

I didn't treat the _API_KEY naming note, to me it already fits how secrets are handled elsewhere. I'll let you tell me if we should do it or not. 🙏 

I also smoke-tested the dual-server case locally:

- Navidrome as one server (username/password)
- K7 / OpenSubsonic as the other (apiKey)

I checked both directions (Navidrome default then K7 default). In the leak scenario, bound password server while a global apiKey from the other server is still in config, requests keep the bound server’s auth only (u/p, no foreign apiKey), and the reverse for the apiKey server. Live test_connection succeeds on both.

image
image
image

NeptuneHub (approved):
I just found some issue, but I directly fixed and pushed against your PR branch, they are:
- credential leak: If credential wasn't saved correctly in the database, it mix credential between the selected server and the default one;
- label issue: here and there you called the api subosnic, instead they are OPENsubosnic.

I approve this change but please re-do on your side a full run of MANUAL/HUMAN test against this last fix to have the 4eye principle. Play a bit around having 2 server.
Also please do a real test on Provider Migration.

If you confirm all of them work I'll proceed with merge. thanks.

kaybi-gh:
I pulled your fixes (thank you 🙏) and re-tested locally.

Dual-server setup:

- Navidrome with username/password
- K7 / OpenSubsonic with apiKey

Checked both directions (Navidrome as default, then K7). In the leak cases:
- bound password server while a global apiKey is still in config
- and the reverse, requests only use the bound server creds

Live test_connection works on both.

Also went through the provider migration path:
- probe Navidrome with password: ok
- probe K7 with apiKey: ok
- started a migration session from Navidrome (password) to K7 (apiKey) and listed libraries on the target: ok
- confirmed the probe does not mix auth methods when the global config still holds the other server's credentials
- wrong password against K7 fails cleanly as auth_failed

So it's looking pretty good on my side.

kaybi-gh:
Thank you again for your time 🙏 🎉

NeptuneHub:
Fix released in the new v3.2.0 here:
- https://github.com/NeptuneHub/AudioMuse-AI/releases/tag/v3.2.0

Also if you like to stay in contact you can reach me on the Jellify Discord server (a mobile app for Jellyfin) where there is a dedicated AudioMuse-AI channel:
- https://discord.gg/c9Quk7J3d

Issue #838, opened 2026-08-03
[FEATURE] Chromaprint fingerprinting should use multiple workers when available


diswell:
Hi,

Currently backfilling my library with chromaprint fingerprints as I noticed there was a cap on the number of fingerprints generated per run. I have a local worker running on Unraid and a remote worker set up on a network PC with a GPU. Both workers are available and showing in the dashboard. 

I launched Analysis with a much higher cap and I can see the fingerprinting is running for my whole library - however, looks like this is only running on the local worker with the remote worker idle. 

<img width="1123" height="606" alt="Image" src="https://github.com/user-attachments/assets/54644ce1-3dd5-4561-a988-8d15b37852e5" />

Previously, distributed analysis has been working fine across the local and remote workers - not sure why the fingerprinting is restricted to one worker.

NeptuneHub:
Chromaprint backfill on already analyzed song it’s only for migration and go only on one worker. New song analyzed instead go on multiple worker as usual.

In short migration, being a one shot that we will never reuse, I prefered to keep it simple.

wirrunna:
"[2026-08-07 08:16:21] Building artist component projection... (8/8)",
    "[2026-08-07 08:16:26] Calculating Chromaprint fingerprints for 284224 track(s) on server 36ab5a91b5f144dcb9d652a10ea3da42...",
    "[2026-08-07 08:21:26] Calculating Chromaprint fingerprints on server 36ab5a91b5f144dcb9d652a10ea3da42: 1/284224 track(s)...",
    "[2026-08-07 08:26:26] Calculating Chromaprint fingerprints on server 36ab5a91b5f144dcb9d652a10ea3da42: 2/284224 track(s)...",
    "[2026-08-07 08:31:26] Calculating Chromaprint fingerprints on server 36ab5a91b5f144dcb9d652a10ea3da42: 3/284224 track(s)..."
  ],
  "log_storage_info": "Log in DB truncated to last 10 entries. Original length: 36.",
  "message": "Calculating Chromaprint fingerprints on server 36ab5a91b5f144dcb9d652a10ea3da42: 3/284224 track(s)...",
  "status_message": "Calculating Chromaprint fingerprints on server 36ab5a91b5f144dcb9d652a10ea3da42: 3/284224 track(s)..."

Does Chromaprint really take 5 minutes per track ?

NeptuneHub:
[...]

Issue #789, opened 2026-07-23
[QUESTION] Why does Chromaprint running outside of of individual album/song analysis?


nguyentn1197:
As asked on the title. why does the chromaprint being the last steps on the analysis?
Looking at the code of https://github.com/NeptuneHub/AudioMuse-AI/blob/6a90277e9669bfaa90ec1ea7cc2debcc91a230ad/tasks/analysis/main.py#L134 
it is basically do the following per track
- Redownload the file to temp storage
- Run the tasks
- Save the result
- Delete the file from temp storage

Why not do that here https://github.com/NeptuneHub/AudioMuse-AI/blob/6a90277e9669bfaa90ec1ea7cc2debcc91a230ad/tasks/analysis/album.py#L224 so it doesn't need to redownload the whole thing again and leverage task distribution to workers to spread out the load and have task tracking?

In my case i have 20000+ songs and the update to v3 cause just the chromaprint to run more than 2 days now without stopping and i don't know the progress

NeptuneHub:
Hi, thanks for the question.

Chromaprint is not really outside the analysis, it depends if the track is new or old.

For **new tracks** it is done exactly where you suggest: `_analyze_single_track` in `tasks/analysis/album.py` calls `_stage_collect_chromaprint` at line 241. At that point the file is already downloaded for the Musicnn analysis, so the fingerprint is calculated on the same temp file, with no extra download, and it is distributed to the workers together with the rest of the album analysis.

The `_backfill_one_track` path in `tasks/analysis/main.py` is only for the **old tracks**, the ones already analyzed before Chromaprint existed. For those the audio file is not on disk anymore (it was deleted right after the analysis), so the only way to add the fingerprint is to download it again. This is why after the update to v3 you see it as a separate step at the end.

This step is also limited on purpose, with this parameter:

`CHROMAPRINT_BACKFILL_ALBUMS_PER_RUN` (default: 1000)

Be careful that it is 1000 **albums** per run, not 1000 tracks: the query takes 1000 albums and then all the tracks of those albums that are missing the fingerprint, so the real number of tracks is much higher. The idea was that every run takes more or less 1-2 hours and the backfill goes ahead a bit at every analysis, so a big library is completed after 1-2 weeks, instead of one huge blocking job.

So in your case with 20000+ songs you can:

- just leave it running, it will complete by itself run after run;
[...]

PR #886, opened 2026-09-02
Optional distributed Chromaprint backfill across worker pool

seitzbg:
## AS-IS

The Chromaprint fingerprint backfill (`tasks/analysis/main.py`) runs as a serial,
in-process `for` loop on whichever single worker executed the `MainAnalysisTask`.
It does not use the task queue: for every album still missing fingerprints it
downloads each track from the media server and runs `fpcalc` one track at a time,
sequentially. On a multi-worker deployment the rest of the worker pool sits idle
through the whole backfill, which on a large library is hours of work.

Album analysis, by contrast, already fans out: it enqueues one
`analyze_album_task` per album via `taskqueue.enqueue(...)` and drains them with
`taskqueue.reap_finished_children(...)`, so every worker helps.

Two closed issues asked for exactly this parallelism: #838 (*"Chromaprint
fingerprinting should use multiple workers when available"*) and #789 (*"Why does
Chromaprint run outside of individual album/song analysis?"*). #838 was declined
to avoid risking bugs in the analysis system for a one-time migration.

## TO-BE

An **opt-in, default-off** distributed backfill that reuses the same proven
enqueue/reap primitives album analysis already uses, so it adds capability without
changing default behavior or touching the album path.

- `CHROMAPRINT_BACKFILL_DISTRIBUTED` (default `false`): when off, the backfill runs
  the existing serial loop **verbatim** (no behavior change). When on, the missing
  fingerprint targets are chunked and enqueued as `chromaprint_backfill` child jobs
  that any idle worker drains, mirroring the album fan-out; the parent reaps the
  finished batches and reports progress in the **same** log format so the UI is
  unchanged. In-flight batches are capped at `MAX_QUEUED_ANALYSIS_JOBS` (the same
  throttle album dispatch uses) so a large library cannot flood the queue in one
  burst. Revocation stops enqueuing and stops waiting, exactly as today.
- `CHROMAPRINT_BACKFILL_BATCH_SIZE` (default `50`, floored at 1): tracks per
  distributed job. Larger batches mean fewer queue rows but coarser progress and
  cancellation; smaller batches spread work more evenly across workers.

A new `chromaprint_backfill_batch_task` worker job binds the same media-server
context the serial loop uses and calls the **unchanged** `_backfill_one_track`
(download then `fpcalc`) for each track in its batch, tolerating a per-track failure
without aborting the batch and publishing its filled/total count for the parent to
aggregate. It is registered in `taskqueue.ALLOWED_FUNCS` and the `tasks.analysis`
facade so it resolves by dotted path like `analyze_album_task`.

**Per-track work is identical either way** - a track is still downloaded from the
media server and fingerprinted with `fpcalc`. Only the scheduling changes. Reducing
the per-track download/fingerprint cost itself is out of scope for this PR.

If distribution is enabled but there is no claimed parent task (a non-queued,
in-process run with no worker pool to help), it falls through to the serial loop.

### Config

| Variable | Meaning | Default |
| :-- | :-- | :-- |
| `CHROMAPRINT_BACKFILL_DISTRIBUTED` | Off = today's serial in-process loop. On = fan the backfill out as `chromaprint_backfill` jobs any worker drains. | `false` |
| `CHROMAPRINT_BACKFILL_BATCH_SIZE` | Tracks per distributed job (only used when distributed). Floored at 1. | `50` |

Documented in `docs/PARAMETERS.md` next to the other `CHROMAPRINT_*` variables.

## Test

Unit tests added in `test/unit/test_analysis.py` (matching the existing backfill
test conventions), covering both paths:

- `test_distributed_backfill_enqueues_one_batch_per_chunk` - flag on, N targets:
  `taskqueue.enqueue` is called ceil(N/batch) times with `task_type='chromaprint_backfill'`
  and the batches partition the targets correctly (mocked enqueue + reaper).
- `test_distributed_backfill_stops_enqueuing_once_revoked` - revocation stops
  enqueuing further batches and returns stopped.
- `test_flag_off_backfill_runs_serially_and_never_enqueues` - flag off: the serial
  per-track loop runs and nothing is enqueued.
- `test_chromaprint_backfill_batch_task_fingerprints_each_track_and_reports_filled`
  - the batch task calls `_backfill_one_track` once per id and reports the filled count.
- `test_chromaprint_backfill_batch_task_keeps_going_when_a_track_fails` - a single
  failing track does not abort the batch.

The 5 pre-existing serial-path backfill tests still pass unchanged, and
`test_taskqueue_registry.py` confirms the new dotted-path job resolves and is callable.

```
$ pytest test/unit/test_analysis.py -k "backfill or distributed or batch_task" test/unit/test_taskqueue_registry.py
test/unit/test_analysis.py::test_chromaprint_backfill_writes_progress_during_the_loop_not_only_before_it PASSED
test/unit/test_analysis.py::test_chromaprint_backfill_throttles_progress_writes_to_the_configured_interval PASSED
test/unit/test_analysis.py::test_chromaprint_backfill_stops_mid_loop_once_the_run_is_revoked PASSED
test/unit/test_analysis.py::test_chromaprint_backfill_leaves_remaining_servers_untouched_once_revoked PASSED
test/unit/test_analysis.py::test_chromaprint_backfill_covers_every_server_when_nothing_is_revoked PASSED
test/unit/test_analysis.py::test_distributed_backfill_enqueues_one_batch_per_chunk PASSED
test/unit/test_analysis.py::test_distributed_backfill_stops_enqueuing_once_revoked PASSED
test/unit/test_analysis.py::test_flag_off_backfill_runs_serially_and_never_enqueues PASSED
test/unit/test_analysis.py::test_chromaprint_backfill_batch_task_fingerprints_each_track_and_reports_filled PASSED
test/unit/test_analysis.py::test_chromaprint_backfill_batch_task_keeps_going_when_a_track_fails PASSED
test/unit/test_taskqueue_registry.py::test_every_allowed_function_imports_and_is_callable[tasks.analysis.chromaprint_backfill_batch_task] PASSED
test/unit/test_taskqueue_registry.py::test_every_allowed_function_imports_and_is_callable[tasks.clustering.run_clustering_batch_task] PASSED
====================== 12 passed ======================
```

Full local run: entire unit suite green (`pytest test/unit/` -> 4103 passed, Python
3.12), and the lint gates pass (`ruff` per `ruff.toml`, `flake8` CI selection,
`codespell`, `mypy` per `mypy.ini`).

## Other useful information

- The change is purely additive and gated behind the opt-in flag; with the flag off
  the serial loop is byte-for-byte the same, so existing deployments are unaffected.
- Not tested on live hardware / against a specific media server in this PR - the
  media-server and `fpcalc` interaction is unchanged (same `_backfill_one_track`),
  only the scheduling around it is new. Opened as a draft for maintainer review of
  the approach before wider testing.

## Checklist

**Type of change:**
- [x] New feature

**Tested on architecture:**
- [ ] Intel
- [ ] ARM
- [ ] -noavx2
- [ ] NVIDIA image

**Tested on media server:**
- [ ] Navidrome
- [ ] Jellyfin
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [x] Documentation
- [x] Unit test
- [ ] Integration test

**Other:**

**Related ISSUE:** Refs #838, #789

## Known tradeoffs (by design)

- **Coarser progress than the serial loop.** The `done/total` counter advances per completed batch rather than per track, and a failed batch does not contribute to `done`, so the reported count can sit slightly below `total` on partial failures. This is cosmetic — the drain still exits on every terminal child state (SUCCESS or FAILURE).
- **Needs more than one worker.** The parent blocks in the drain loop while the batch jobs wait on the default queue for *other* workers to claim them, so distribution only helps a multi-worker deployment; on a single worker, leave `CHROMAPRINT_BACKFILL_DISTRIBUTED` off (the serial path). This matches how album fan-out already behaves.


NeptuneHub:
Hi and thanks for your effort. Here we are talking of a one time migration, needed around 1 months and half ago, designed to don’t be not to much invasive because is something that user will use only one time in their life.

I know that the backfill in this way could take a week or so depending from library size and hw, but it’s also true that in the meanwhile you can just keep using AudioMuse-AI and keeping the backfill in the background.

I also closed the linked issue exactly for this reason.

so I’m going to reject this PR because I feel that the regression risk is higher and more time expensive from a testing/fix time than the value that it could eventually bring to the remaining user that still didn’t migrated.

rawsocket-dev:
Ah makes sense, thanks!

PR #906, opened 2026-09-12
Expose authenticated model coverage and application version APIs

rendyhd:
## AS-IS

Ordinary clients cannot read numeric coverage for all models. The setup wizard exposes admin-only visual bands, and application version reporting is not available through a dedicated authenticated route.

## TO-BE

Adds two endpoints available to ordinary authenticated users and bearer-token clients:

- `GET /api/models`: effective enablement and global coverage for MusiCNN (`musicnn`), DCLAP (`clap`), Lyrics (`lyrics`), and Neural Fingerprint (`neural-fingerprint`). Each coverage object contains `count`, `total`, and `percentage`.
- `GET /api/models?server_id=<id-or-name>` (also `server=`): keeps global coverage and adds local coverage for the resolved server. Without a server parameter, local coverage and `server_id` are omitted.
- `GET /api/version` in `app.py`: returns only the effective `app_version`, including release suffixes.

Model coverage contains no API/application version or recording-limit fields. MusiCNN is always enabled, matching the wizard. Percentages use the whole relevant catalogue for every model, including Lyrics. Unknown index counts remain null, empty catalogues have zero percent, and percentages are bounded to 0-100 without masking stale raw counts.

Shares the existing wizard count sources through `tasks/model_coverage.py`; the wizard's bands and lyrics-eligible denominator remain unchanged. Reuses source-scoped neural counting and adds cached paged-IVF directory/mapping counts. Local queries never substitute global coverage. The enablement/model-file-availability split remains intact.

The endpoints preserve the existing auth/setup policy, return sanitized errors and `Cache-Control: no-store`, and do not load encoders/search indexes, infer, analyze, or renew recording warmup. A cold local paged-IVF check reads compact directory IDs and source mappings, not cells/embeddings; scalar counts are cached for up to 30 seconds and invalidated on mapping/index replacement.

Includes API schemas and `docs/model-coverage-api.md`. No migration, dependency or Lumae frontend change.

## Test

Local Windows x64 / Python 3.14.2 validation:

- 164 passed, 1 skipped in model API, wizard, paged/neural index, availability, recording, header and import-architecture suites.
- After adding generated-OpenAPI validation: 146 passed in model API, authentication, shared coverage and header suites (overlaps the preceding run).
- Repository-wide Ruff, flake8, codespell and scoped mypy passed; generated OpenAPI includes both endpoints; tracked-file header checks and `git diff --check` passed.

Reproduce the relevant unit checks:

```sh
python -m pytest test/unit/test_model_coverage_api.py test/unit/test_model_coverage.py test/unit/test_setup_wizard_model_coverage.py test/unit/test_setup_wizard_model_switches.py test/unit/test_app_recording_search.py test/unit/test_neural_fingerprint.py test/unit/test_neural_fingerprint_index.py test/unit/test_paged_ivf.py test/unit/test_index_availability_cache.py test/unit/test_file_header_convention.py test/unit/test_import_architecture.py test/unit/test_app_auth.py -q
```

New real-PostgreSQL integration regression:

```sh
python -m pytest test/integration/test_model_coverage_api_integration.py -q
```

It verifies persisted directory/header counts, default legacy IDs, scoped canonical mappings, lyrics percentages and an emptied secondary source using an isolated schema in the disposable test database. It skips locally because no disposable PostgreSQL fixture is installed; review CI for the full unit/integration results. No live provider, ARM/NVIDIA or 150k-song deployment was tested locally.

## Other useful information

Follows the maintainer's requested all-model API shape: https://github.com/NeptuneHub/AudioMuse-AI/pull/906#issuecomment-5653738139

Implementation and this description were prepared with Codex assistance.

## Checklist

**Type of change:**
- [x] New feature
- [x] Documentation
- [ ] Breaking change

**Tested on media server:**
- [ ] Navidrome
- [ ] Jellyfin
- [ ] Emby
- [ ] Lyrion

**Updated:**
- [x] Documentation
- [x] Unit test
- [x] Integration test

**Other:**

### DIFF
--- a/tasks/neural_fingerprint_index.py
+++ b/tasks/neural_fingerprint_index.py
@@ -155,6 +155,7 @@ PICKER_WHERE = (
 OLDER_LAYOUT = 'The neural fingerprint index was built by an older version; the next analysis run rebuilds it.'
 LOAD_FAILED = 'The neural fingerprint index could not be loaded; check the container logs.'
 _AVAILABILITY = AvailabilityCache()
+_STATUS_COUNTS = AvailabilityCache()
 _CANONICAL = {}
 _LOCK = threading.RLock()
 _STATE = {'pack': None, 'building': False, 'error': None, 'gpu': None, 'executor': None}
@@ -885,6 +886,40 @@ def get_status():
         }
 
 
+def get_scoped_status(server_id):
+    """Read one loaded index snapshot without loading models or renewing warmup.
+
+    Counts use the search availability rules, cached per server/build for 30 s
+    and invalidated with mapping changes and pack swaps. Never fall back to the
+    union catalogue if source resolution or availability lookup fails.
+    """
+    if not server_id:
+        raise RuntimeError('A resolved music server is required for index status')
+    with _LOCK:
+        pack = _STATE['pack']
+        building = _STATE['building']
+        error = _STATE['error']
+    if building:
+        return {'state': 'loading', 'indexed_tracks': None}
+    if error:
+        return {'state': 'error', 'indexed_tracks': None}
+    if pack is None:
+        return {'state': 'not_loaded', 'indexed_tracks': None}
+
+    def count_available():
+        if not pack.live_tracks or _mask_unneeded(pack, server_id):
+            return pack.live_tracks
+        from database import get_db
+
+        mask = build_availability_mask(server_id, pack.ids, get_db)
+        if mask is None:
+            raise RuntimeError('Selected-server availability could not be determined')
+        return int(np.count_nonzero(mask))
+
+    count = _STATUS_COUNTS.get(server_id, pack.build_id, count_available)
+    return {'state': 'ready' if count else 'empty', 'indexed_tracks': count}
+
+
 def _db_connection():
     from flask import has_app_context
 
@@ -1088,6 +1123,7 @@ def identify(audio, sr, n_results):
 
 def invalidate_availability_cache(server_id=None):
     _AVAILABILITY.invalidate(server_id)
+    _STATUS_COUNTS.invalidate(server_id)
 
 
 def _has_canonical_ids(pack):
--- a/app_recording_search.py
+++ b/app_recording_search.py
@@ -17,7 +17,8 @@ Main Features:
 * Routes: `/recording_search` page, `/api/recording_search/search` (multipart:
   the clip and a result count), `/api/recording_search/by_track` (JSON: a
   library song's id and a result count, for the other recordings of that
-  song) and `/api/recording_search/warmup`.
+  song), `/api/recording_search/warmup`, and read-only
+  `/api/recording_search/status` (ordinary authenticated users).
 * The upload is handed to the manager as a stream, never read into memory
   here; a Content-Length past RECORDING_SEARCH_MAX_UPLOAD_MB answers 413 before
   any byte is copied.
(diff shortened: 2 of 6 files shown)


NeptuneHub:
Hi, thanks a lot for the work here! Before we go further I'd like to change the shape of this API, so let me explain what I have in mind.

What I need is **one endpoint that any logged-in user can call** (not admin only). For **each model** it should return:

- **enabled**: on / off
- **global coverage**: coverage over the whole catalogue, as a number and a percentage
- **local coverage**: coverage over one specific server, only when a server is passed in the request, again as a number and a percentage

So it should cover all the models we show in the setup wizard (MusiCNN, DCLAP, Lyrics, Neural Fingerprint), not only the neural fingerprint one. At the moment the PR covers one model out of four, and nothing in the frontend calls the new route yet, so a normal user still doesn't see anything new.

**Good news:** most of the logic already exists. `model_coverage_levels()` in `app_setup.py` already computes coverage for all four models, and your `get_scoped_status()` already does the per-server counting. I think you can build the new endpoint mostly by reusing those two.

**One more thing:** please remove the version and the recording limits from this response. `api_version`, `app_version` and the whole `recording` block are a different topic, and mixing them here will make the API harder to keep stable.

If the client really needs the server version, let's not put it here. Let's create a `GET /api/version` in `app.py`, still gated from the fact that user need to be authenticated.

The rest is good and I'd like to keep it: the `is_available()` / `model_files_available()` split, the auth approach, and the tests. Please carry them over.

Thanks again!

rendyhd:
I've made the changes. Version is very useful when you want to integrate into an app. Now I can put a message up: In order to use you have to update to version

rendyhd:
after merging main tests now fail, i'll fix

tasks/neural_fingerprint_index.py
```diff
@@ -885,6 +886,40 @@ def get_status():
         }
 
 
+def get_scoped_status(server_id):
```
NeptuneHub: **`get_scoped_status()` can return `null` for a loaded index, because `_STATE['error']` can be stale.**

**The flow (fresh install):**

1. The web process starts. `load_at_startup()` → `ensure_loaded()` tries to load the index, but it is not built yet. The load fails and `_record_load_error()` stores "No neural fingerprint index is built yet" in `_STATE['error']`. `_STATE['pack']` stays `None`. This is correct at this moment.
2. The user runs the analysis. The worker builds the index, and the reload listener in `app.py` calls `reload_from_db()` → `_swap_pack()`, which sets `_STATE['pack']`.
3. Nothing clears `_STATE['error']`. It is only reset inside `ensure_loaded()` when a load starts from the unloaded state, and that never runs again, because `ensure_loaded()` returns early once a pack is loaded.

**Why this is a problem:** after step 2 the state is `pack = loaded` and `error = "not built yet"` at the same time. The error describes the past, not the present. `get_scoped_status()` checks `error` before `pack`, so it stops at the old message and returns `indexed_tracks: None`. The result is that `/api/models?server_id=...` returns a global neural-fingerprint count while `local.count` is `null` until the container restarts. The same happens for any startup load failure that is later followed by a successful reload.

**Why it was not visible before:** the existing readers are not affected. The Setup Wizard uses `indexed_track_count()`, which never reads `error`, and `_pack_state()` checks `loaded` first. `get_scoped_status()` is the first reader that trusts `error` before the pack.

Please make both changes:

**1. Clear the stale error at the source, in `_swap_pack()`.** A successfully swapped pack means the previous load error is no longer true.

```python
def _swap_pack(pack):
    with _LOCK:
        _STATE['pack'] = pack
        _STATE['error'] = None
```

**2. In `get_scoped_status()`, check the pack first.** Replace the `if building:` / `if error:` / `if pack is None:` checks with the following, so a loaded pack always wins. This is the same precedence as the existing `_pack_state()` in `tasks/recording_search_manager.py` (loaded → building → error), and it keeps the function correct even if an error is recorded while a pack is loaded.

```python
    if pack is None:
        if building:
            return {'state': 'loading', 'indexed_tracks': None}
        if error:
            return {'state': 'error', 'indexed_tracks': None}
        return {'state': 'not_loaded', 'indexed_tracks': None}
```

**Tests:** the two `loaded=True` rows of `test_neural_unknown_coverage_stays_null` (`True-True-None-loading` and `True-False-secret-error`) currently assert the wrong behaviour. With a loaded pack they should expect the real count. Please also add a regression test for the flow above: `ensure_loaded()` with no stored index, then `reload_from_db()` → pack loaded, error cleared, local count returned.

One change of mind on my side, sorry for the back and forth. Your first version used `selected_server_scope()` (no server passed → default server). Then I asked for local coverage "only when a server is passed", and you correctly removed the fallback. Thinking about it again, I prefer the standard behaviour of the other endpoints, as in your first version.

Please use `app_server_context.selected_server_scope()` in the route again, instead of `resolve_request_server_id()` plus the empty-parameter check. It returns `(server_id, is_default)`, so `get_model_coverage()` can receive `is_default` as `include_legacy` instead of comparing with `registry.get_default_server_id()` again.

This means:
- `local` and `server_id` are always present (unless no server is configured).
- An empty `?server=` / `?server_id=` falls back to the default instead of returning 400, like everywhere else.
- Tests to update: `test_global_only_contract_uses_whole_catalogue_even_for_lyrics` and the two empty-value rows of `test_invalid_scope`. Please also update the Swagger text and `docs/model-coverage-api.md`.

